PR #13, opened 2019-02-01
Apple ASR and Continued work on CoreML WakeWord

kwylez:
@noelweichbrodt Not sure why this can't be merged, but this PR contains improvements to the wakeword CoreML, as well as, a refactored main view controller UI that allows the user to select which service they want to use:
* wakeword (disabled for now)
* google speech to text (need api key)
* apple's asr


SpokeStack/SpokeStack/WakeWordSpeechRecognizer.swift
```diff
@@ -0,0 +1,660 @@
...
+        /// Transfer the mel filterbank window to the detector model's inputs
+        
+        self.frameWindow.rewind()
+        
+        guard let multiArray = try? MLMultiArray(shape: [1,40,40], dataType: .float32) else {
+            
+                fatalError("Unexpected runtime error. MLMultiArray")
+        }
+        
+        for index in 0..<ModelConstants.numOfMelOutputs {
```
brentspell: This loop is not copying the values from the frameWindow into the detector inputs, and you are only filling 40 values in the input array. I suspect this is why you are getting NaN values in your detector output, since the remainder of the array is not being initialized. This array should be filled with the contents of the frame window that was populated by the `filter` method. See https://github.com/pylon/spokestack-android/blob/master/src/main/java/com/pylon/spokestack/wakeword/WakewordTrigger.java#L461 for the correct implementation in the android repo, lines 463-466.

SpokeStack/SpokeStack/WakeWordSpeechRecognizer.swift
```diff
@@ -0,0 +1,660 @@
...
+        self.phrase()
+    }
+    
+    private func phrase() -> Void {
+        
+        /// Compute the argmax (winning class) of each smoothed output
+        /// in the current phrase window
+        
+        var index: Int = 0
+        var max: Float = -Float.greatestFiniteMagnitude
```
brentspell: This max needs to be initialized within the phrase window loop, because we want to calculate the winning word class across for each phrase frame. See https://github.com/pylon/spokestack-android/blob/master/src/main/java/com/pylon/spokestack/wakeword/WakewordTrigger.java#L501.

SpokeStack/SpokeStack/WakeWordSpeechRecognizer.swift
```diff
@@ -0,0 +1,660 @@
...
+            /// Search for any occurrences of the phrase's keywords in order
+            /// across the whole phrase window
+            
+            var match: Int = 0
+            
+            phraseArgumentLabel: for word in self.phraseArg {
+                
+                if word == phrase[match] {
+                    
+                    match -= 1
```
brentspell: I believe this should be an increment: https://github.com/pylon/spokestack-android/blob/master/src/main/java/com/pylon/spokestack/wakeword/WakewordTrigger.java#L519

brentspell:
In general, it looks like there is a significant amount of work that still needs to be done to finish the wakeword trigger port. It isn't raising the spokestack activation/deactivation events yet, configuration is hard-coded to the updog wakeword model, the latest changes from the android version haven't been integrated, and there are no tests.

noelweichbrodt:
I have merged the immediately relevant files from this pr into another pr focused just on apple asr and asr wakeword (and the updated example ui). Closing this with thanks.

PR #32, opened 2019-03-26
Ignore Apple error codes that indicate retries

noelweichbrodt:
Per https://developer.nuance.com/public/Help/DragonMobileSDKReference_iOS/Error-codes.html, there are swaths of `SFSpeechAudioBufferRecognitionRequest` errors that aren't errors, just indications to retry the recognition request. Since we don't store audio for a retry, there is nothing else to do if a request results in one of these errors. So we will simply ignore the error and keep going. Note that the lifespan of the request is not affected by these, so streaming will continue until control flow ends it.


noelweichbrodt:
[Spokestack iOS kAFAssistantErrorDomain Code 1](https://trello.com/c/ItSL9VVk/163-spokestack-ios-kafassistanterrordomain-code-1)

PR #33, opened 2019-03-29
take advantage of context to cut down on unecessary stop() calls

noelweichbrodt:
This should prevent errors like that in the screenshot, which shouldn't ever happen according to Apple.

Screen Shot 2019-03-28 at 4 56 42 PM

Additionally, this is a candidate for fixing the "com.apple.coreaudio.avfaudio error 10851" exception, whose cause still remains unknown.

### DIFF
--- a/SpokeStack/SpeechPipeline.swift
+++ b/SpokeStack/SpeechPipeline.swift
@@ -95,8 +95,10 @@ import Foundation
     }
     
     @objc public func start() -> Void {
-        self.stop()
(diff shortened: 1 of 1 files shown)


noelweichbrodt:
[Spokestack iOS com.apple.coreaudio.avfaudio error 10851](https://trello.com/c/fJrYOIga/191-spokestack-ios-comapplecoreaudioavfaudio-error-10851)

PR #34, opened 2019-04-03
Use iOS AudioSession more sparingly to help with other clients

noelweichbrodt:
In particular, don't end an AudioSession upon deactivation. Leave it open in case other parts of the iOS client are actively playing audio


noelweichbrodt:
[Radiobrain stops playing when put in the background for non-audio apps](https://trello.com/c/uqOeDgtb/200-radiobrain-stops-playing-when-put-in-the-background-for-non-audio-apps)

PR #37, opened 2019-04-12
iOS all outside audio should be silenced upon init

noelweichbrodt:
Removes ducking from ios audiosession categories and don't disable audiosession when setting categories.


noelweichbrodt:
[Radiobrain will play audio while other open apps play audio](https://trello.com/c/rHegJYWK/199-radiobrain-will-play-audio-while-other-open-apps-play-audio)

PR #40, opened 2019-04-24
don't coerce remoteIOUnit

noelweichbrodt:
With the refactor of AudioController, the remoteIOUnit optional property is not always being set. Stop coercing that property during `stop` and `deinit`.


noelweichbrodt:
[[Spokestack] calling stop causes a crash in iOS](https://trello.com/c/iwJbEO8t/267-spokestack-calling-stop-causes-a-crash-in-ios)

PR #42, opened 2019-05-03
move audiosession category check from init to startStreaming

noelweichbrodt:
Since a client may reasonably have the `AudioSession` category set to something else before actually calling `spokestack.start()`, don't throw an error at `AudioController.init()` and instead throw an error at `AudioController.startStreaming()`.


noelweichbrodt:
[Airpods do not reconnect after disconnecting](https://trello.com/c/d4FNaRtu/266-airpods-do-not-reconnect-after-disconnecting)

PR #44, opened 2019-06-06
Voice-limited activity detector

noelweichbrodt:
Implements a VLAD for detecting when a voice is heard and activating the wakeword recognizer. Once a voice is no longer heard, VLAD deactivates the wakeword recognizer. When the spokestack pipeline is active, VLAD does not interrupt.


brentspell:
lgtm, nice that you were able to fix the witvad crashes and get it working. I don't think cory ever got it to stop crashing, which is why he started working on porting the webrtc version.

PR #45, opened 2019-06-14
CoreML wakeword detection

noelweichbrodt:
Based on @kwylez's port of `spokestack-android`'s tensorflow-lite model-based wakeword detector. There is a default model and configuration included, and an example usage for testing.


noelweichbrodt:
[Spokestack iOS integrate CoreML-based wakeword](https://trello.com/c/iPGmYTYY/166-spokestack-ios-integrate-coreml-based-wakeword)

noelweichbrodt:
There is a [heap overflow memory bug](https://github.com/wit-ai/wit-ios/issues/88) in WitVAD but otherwise, this implementation is working. We will port WebRTC's VAD in a future PR and use that instead.

PR #46, opened 2019-08-02
Use (a fork of) WebRTC's VAD

noelweichbrodt:
Deprecate the use of WitVAD and use WebRTC VAD instead. 

Includes a new Swift wrapper for C++ interop with `filter_audio`, along with improvements to `RingBuffer` for buffering between CoreAudio and the VAD, and some `CoreMLWakeword` bug fixes and compatibility updates.

In testing, this has no heap buffer overflows and allows the CPU to idle at ~10% when there's no speech. Running the CoreML wakeword model increases CPU use to ~110%.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "SpokeStack/filter_audio"]
+	path = SpokeStack/filter_audio
+	url = git@github.com:noelweichbrodt/filter_audio.git
```
brentspell: it would be better if this was forked into pylon instead of your personal account

SpokeStack/AppleWakewordRecognizer.swift
```diff
@@ -28,21 +28,19 @@ public class AppleWakewordRecognizer: NSObject, WakewordRecognizerService {
     private var recognitionTask: SFSpeechRecognitionTask?
     private let audioEngine: AVAudioEngine = AVAudioEngine()
     private var dispatchWorker: DispatchWorkItem?
-    private var vad: WITVad = WITVad()
```
brentspell: rip

noelweichbrodt: unmourned

PR #47, opened 2019-08-21
Add support for efficient autoregressive wakeword detection using TensorFlowLite

noelweichbrodt:
Adds support (via `TensorFlowLiteSwift`) for TensorFlowLite model-based wakeword detection, and an example model implementation.

This is a companion PR to https://github.com/pylon/spokestack-android/pull/13, see it for further information.

Readme update, tests, and tracing support will be featured in subsequent PRs.


SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+            let frameWidth: Int = c.frameWidth
+            self.minActive = c.wakeActiveMin / frameWidth
+            self.maxActive = c.wakeActiveMax / frameWidth
+        }
+    }
+    
+    /// MARK: Pipeline control
+    
+    private func process(_ frame: Data) -> Void {
+        self.activeLength += 1
```
brentspell: The logic here is incorrect. The active length increment and max active check should only be done if the pipeline is active: https://github.com/pylon/spokestack-android/blob/master/src/main/java/com/pylon/spokestack/wakeword/WakewordTrigger.java#L357.

It's dangerous, because every `activeLength` frames, you are going to fall into the else side and reset all of the sliding windows, even if the pipeline is not active.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+            self.prevSample = currentSample
+            
+            //sampleCollector.append(sample)
+            
+            /// Process the sample
+            /// - write it to the sample sliding window
+            /// - run the remainder of the detection pipleline if speech
+            /// - advance the sample sliding window
+            try self.sampleWindow.write(sample)
+            if self.sampleWindow.isFull {
```
brentspell: This is problematic because returning here will ignore any remaining audio samples in `dataElements` once the sample window fills. This is why the android version calls the subsequent functions inside the sample loop.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+                return
+            }
+        }
+    }
+    
+    private func analyze() throws -> Void {
+        /// The current sample window contains speech, so
+        /// apply the fft windowing function to it
+        if self.sampleWindow.isFull {
```
brentspell: This check is not needed if this function is called in the sample loop.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+            self.sampleWindow.rewind().seek(self.hopLength)
+            
+            //self.fftFrameCollector += "\(self.fftFrame)\n"
+        }
+    }
+    
+    /// Attention model processing
+    
+    private func filter() throws -> Void {
+        if !self.fftFrame.isEmpty {
```
brentspell: This check is not needed if filter is called from analyze.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+                }
+                
+            } catch let message {
+                throw WakewordModelError.filter("TFLiteWakewordRecognizer filter \(message)")
+            }
+        }
+    }
+    
+    private func encode() throws -> Void {
+        if self.frameWindow.isFull {
```
brentspell: Same here.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+                    self.frameWindow.rewind()
+                    // TODO: model.copy requires that the data be sized to exactly the same as the tensor, so we can't just do read()s off the ringbuffer and copy over piecewise. This introduces an aggrevating overhead of having to copy the ringbuffer into an array before copying over to the tensor. Maybe use a fixed-sized array that is advanced based off the fft frame size?
+                    var frameWindowArray: Array<Float32> = []
+                    while !self.frameWindow.isEmpty {
+                        let f = try self.frameWindow.read()
+                        frameWindowArray.append(f)
+                    }
+                    var stateArray: Array<Float32> = []
+                    for _ in 0..<self.stateWidth {
+                        let f = try self.encodeState.read()
```
brentspell: I was thinking that somebody would need to reset the `encodeState` ring buffer, but it is neatly reset every time it is read/written. Nice.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+            do {
+                if let model = self.detectModel {
+                    /// inputs
+                    var encodeWindowArray: Array<Float32> = []
+                    self.encodeWindow.rewind()
+                    while !self.encodeWindow.isEmpty {
+                        let f = try self.encodeWindow.read()
+                        encodeWindowArray.append(f)
+                    }
+                    _ = try encodeWindowArray
```
brentspell: It would probably be a good idea to create a simple swift abstraction around these tensor reads/writes to simplify the code (later). Those abstractions could also be used to avoid the temporary array allocations inside the frame loop.

SpokeStack/TFLiteWakewordRecognizer.swift
```diff
@@ -0,0 +1,478 @@
...
+    }
+    
+    private func reset() -> Void {
+        /// Empty the sample buffer, so that only contiguous
+        /// speech samples are written to it
+        self.sampleWindow.reset()
+        
+        /// Reset and fill the other buffers,
+        /// which prevents them from lagging the detection
+        self.frameWindow.reset().fill(0)
```
brentspell: The other 3 ring buffers (encode, state, and detect) should also be reset/filled here.

noelweichbrodt:
These are good points, thanks @brentspell!

PR #49, opened 2019-08-30
Tracing event threaded through pipeline

noelweichbrodt:
Creates a  new `didTrace` event with variable-level logging that propagates debugging events back to the client. This approach also resulted in additional advantageous simplifications in class/protocol structure.


noelweichbrodt:
[Spokestack iOS enable tracing levels](https://trello.com/c/jmaW1D52/167-spokestack-ios-enable-tracing-levels)

brentspell:
lgtm

PR #53, opened 2019-10-09
Dynamically-loaded CoreML models

noelweichbrodt:
Move CoreML wakeword models into example project and load them dynamically in Spokestack. This matches the existing behavior of the TensorFlowLite models.
Also:
- Tidy up the organization of `CoreMLWakewordRecognizer.swift`
- Harmonize the names of both CoreML and TFLite models.

Note: the pre-compiled `Detect.swift` and `Filter.swift` are derived from their corresponding `mlmodel` files. If the `mlmodel`s change, then these will need to be regenerated.


noelweichbrodt:
[spokestack-ios: react-native interface](https://trello.com/c/NfuyYK7Z/443-spokestack-ios-react-native-interface)

PR #56, opened 2019-11-01
Update badge image

space-pope:


noelweichbrodt:
Weird, I swear I changed the badge. Thanks for catching it!

space-pope:
The link _after_ the badge was right, so if you clicked on it, you'd go to the right page, but the link for the badge image itself pointed to the old SVG.

PR #62, opened 2019-11-20
Text to Speech API

noelweichbrodt:
A simple `TextToSpeech` API with a single `synthesize` function. This function calls the Spokestack TTS API service and returns a URL to a streaming MP3 file (or an error) to the registered TTS delegate. Additionally, this new TTS API includes:
  * Integration tests for the `synthesize` function
  * Swift Markdown documentation
  * TTS API tracing for debugging request/response
  * Reference usage in the `SpokestackFrameworkExample` app


Spokestack/TextToSpeech.swift
```diff
@@ -0,0 +1,79 @@
...
+    private var delegate: TextToSpeechDelegate
+    private var configuration: SpeechConfiguration
+    
+    /// Initializes a new text to speech instance.
+    /// - Parameter delegate: Delegate that receives text to speech events.
+    /// - Parameter configuration: Speech configuration parameters.
+    init(_ delegate: TextToSpeechDelegate, configuration: SpeechConfiguration) {
+        self.delegate = delegate
+        self.configuration = configuration
+        super.init()
```
kwylez: @noelweichbrodt Is there a necessity to have a `strong` reference to the `delegate`? Should that be a `public` scope?

Something like: `weak public var delegate: TextToSpeechDelegate?`

noelweichbrodt: good point!

PR #63, opened 2019-11-21
Fix TextToSpeechInput.inputFormat; SpeechPipeline.setDelegate

noelweichbrodt:
Both of these are currently crippled due to authorial oversight. This fixes them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,15 @@ Activity Detection (VAD), wakeword activation, and Automatic Speech Recognition 
 
 This example creates a speech recognition pipeline using a wakeword detector that is triggered by VAD, which in turn activates an ASR, returning the resulting utterance to the `SpeechEventListener` event listener (`self` in this example).
 
+See `SpeechPipeline` and `SpeechConfiguration` for further configuration documentation.
+
 ### Reference implementation
 
-The `SpokestackFrameworkExample` project contains reference implementations for how to use the Spokestack library, along with runnable examples of the wakeword and ASR components.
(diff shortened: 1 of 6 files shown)


Spokestack/TextToSpeech.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-@objc public enum TTSInputType: Int {
+@objc public enum TTSInputFormat: Int {
```
space-pope: If this were `TTSInputFormat: String`, you could use `input.inputFormat.rawValue` below and do away with the `switch`.

README.md
```diff
@@ -45,11 +45,15 @@ Activity Detection (VAD), wakeword activation, and Automatic Speech Recognition
...
+
 ### Reference implementation
 
-The `SpokestackFrameworkExample` project contains reference implementations for how to use the Spokestack library, along with runnable examples of the wakeword and ASR components.
+The `SpokestackFrameworkExample` project contains reference implementations for how to use the Spokestack library, along with runnable examples of the wakeword and ASR components. Each component has a corresponding screen from the main screen, and can be started, stopped, or synthesized, as appropriate. The component screens have full debug tracing enabled, so the system control logic and debug events will appear in the XCode Console.
 
-See `SpeechPipeline` and `SpeechConfiguration` for further configuration documentation.
+#### Troubleshooting
+
+A build error similar to `Code Sign error: No unexpired provisioning profiles found that contain any of the keychain's signing certificates` will occur if the bundle identifier is not changed from `io.Spokestack.SpokestackFrameworkExample`, which is tied to the Spokestack organization.
```
space-pope: This wasn't my exact error text—it could probably be generalized to "a code signing error", but not a big deal.

Spokestack/TextToSpeech.swift
noelweichbrodt: objc interop requires `Int` :````````(

PR #66, opened 2020-01-02
Revised TTS API

noelweichbrodt:
Revised TTS API to use new HMAC+GraphQL service. This new service required a couple of deprecations to the existing library API, namely the new `TextToSpeechResult` class replaces `URL`, and no longer supporting plain-text `TextToSpeechInput`. HMAC support requires iOS 13+. The new service supports an optional client-supplied request identifier, which is echoed in the result.

### DIFF
--- a/Spokestack/TextToSpeech.swift
+++ b/Spokestack/TextToSpeech.swift
@@ -8,12 +8,7 @@
 
 import Foundation
 import AVFoundation
-
-@objc public enum TTSInputFormat: Int {
-    case text
-    case ssml
-}
-
+import CryptoKit
 
 /**
  This is the client entry point for the Spokestack Text to Speech (TTS) system. It provides the capability to synthesize textual input, and speak back the synthesis as audio system output. The synthesis and speech occur on asynchronous blocks so as to not block the client while it performs network and audio system activities.
@@ -30,6 +25,7 @@ import AVFoundation
  tts.speak(input) // synthesize the same input as above, and play back the result using the default audio system.
  ```
  */
+@available(iOS 13.0, *)
 @objc public class TextToSpeech: NSObject {
     
     // MARK: Properties
@@ -39,6 +35,8 @@ import AVFoundation
     
     private var configuration: SpeechConfiguration
     private lazy var player: AVPlayer = AVPlayer()
+    private var apiKey: SymmetricKey?
+    private let ttsInputVoices = [0: "demo-male"]
     
     // MARK: Initializers
     
@@ -48,6 +46,12 @@ import AVFoundation
     @objc public init(_ delegate: TextToSpeechDelegate, configuration: SpeechConfiguration) {
         self.delegate = delegate
         self.configuration = configuration
+        // create a symmetric key using the configured api key
+        if let apiKeyEncoded = self.configuration.apiKey.data(using: .utf8) {
+            self.apiKey = SymmetricKey(data: apiKeyEncoded)
+        } else {
+            self.delegate?.failure(error: TextToSpeechErrors.apiKey("Unable to encode apiKey."))
+        }
         super.init()
     }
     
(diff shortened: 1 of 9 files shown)


Spokestack/TextToSpeech.swift
```diff
@@ -172,3 +184,15 @@ import AVFoundation
         }
     }
 }
+
+fileprivate struct SynthesizeResponseURL: Codable {
```
space-pope: Is this one ever used?

Spokestack/TextToSpeech.swift
```diff
@@ -172,3 +184,15 @@ import AVFoundation
...
+    let url: URL
+}
+
+fileprivate struct SynthesizeResponseSynthesize: Codable {
```
space-pope: This name's pretty confusing.

Spokestack/TextToSpeechInput.swift
```diff
@@ -8,46 +8,47 @@
 
 import Foundation
 
+/// Designate the format the input is provided in.
+@objc public enum TTSInputFormat: Int {
+    @available(*, unavailable, message: "Text-only input is not currently supported. Encase your text with <speak>text</speak> for SSML compatibility.")
```
space-pope: I think we'll have API support for plaintext soon enough that we can reverse this deprecation and not merge until it exists.

Spokestack/TextToSpeechInput.swift
```diff
@@ -8,46 +8,47 @@
...
-        self.inputFormat = inputFormat
-    }
-    
-    /// Convenience initializer for a new TextToSpeechInput instance.
     /// - Parameter input: The text input to the speech synthesizer.
     /// - Parameter voice: The synthetic voice used to generate speech.
     /// - Parameter inputFormat: The formatting of the input.
-    @objc public init(_ input:String, voice: String, inputFormat: TTSInputFormat) {
+    /// - Parameter id: A unique identifier for this input request.
+    @objc public init(_ input:String = "<speak>Here I am, a brain the size of a planet.</speak>",
```
space-pope: A default value for the text to synthesize seems unnecessary or maybe even counterproductive—or is there an xcode-related reason for this?

Spokestack/TextToSpeechInput.swift
```diff
@@ -8,46 +8,47 @@
...
+        super.init()
     }
     
     /// The synthetic voice used to generate speech.
-    @objc public var voice: String = "demo-male"
+    @objc public var voice: TTSInputVoice = .demoMale
     /// The input to the synthetic voice.
     /// - Note: SSML should be unescaped.
-    @objc public var input: String = "Here I am, a brain the size of a planet."
+    @objc public var input: String = "<speak>Here I am, a brain the size of a planet.</speak>"
```
space-pope: "should be unescaped"? We're not escaping XML anywhere, are we? A better way to put this might be "SSML must be valid XML"—in other words, nested tags don't get escaped, but things like bare `<` and `&` have to be turned into HTML entities.

Spokestack/TextToSpeech.swift
```diff
@@ -81,28 +89,37 @@ import AVFoundation
...
-        var inputFormat: String
-        switch input.inputFormat {
-        case .text:
-            inputFormat = "text"
-            break
-        case .ssml:
-            inputFormat = "ssml"
-            break
+        let body: [String:Any] = [
+            "query":"query synthesis($voice: String!, $ssml: String!) {synthesize(voice: $voice, ssml: $ssml) {url}}",
```
space-pope: @NeilMenne asked me to have a platform-specific name for my GraphQL query to ease auditing. Do you want them to be parallel, Neil, or just have Android queries call out their platform?

noelweichbrodt:
> I do wish there were a better way to do that series of guards for response unwrapping—I thought the structs were supposed to help with that
This is the structs helping! The commit log holds much worse :`(

Spokestack/TextToSpeech.swift
noelweichbrodt: yes, by the struct immediately below it

Spokestack/TextToSpeech.swift
noelweichbrodt: I don't pick the response structure, and the struct naming scheme seems reasonable 🤷‍♂. Open to suggestions though!

Spokestack/TextToSpeechInput.swift
noelweichbrodt: Because this has to be a class for `@objc` compatibility, it must either be initialized or optional. in cases like this, setting a default value allows for both a simpler implementation and a no-configuration-necessary client use case. In scenarios where the client does not directly use a property, I use optionals.

Spokestack/TextToSpeech.swift
noelweichbrodt: I can do whatever for the query name.

Spokestack/TextToSpeech.swift
space-pope: Weird; I saw that relationship in the second and third, but my brain didn't register this one for some reason.

Spokestack/TextToSpeech.swift
space-pope: What about `TTSResponse` `Body`/`Method`/`URL`?

Spokestack/TextToSpeechInput.swift
space-pope: So just including it in the initializer without a default value doesn't count as it being initialized?

noelweichbrodt: well, i am surprised to find out that class properties indeed do not require default values if they are initialized in the only provided constructor.

PR #67, opened 2020-01-07
Cleanup to remove deprecated and unused API interfaces

noelweichbrodt:
These are breaking changes, but only to things that have either been marked as deprecated or are unused parts of a public interface. Realistically, these should not affect any actual clients.


noelweichbrodt:
> This all goes under the current pending major version release, right?

Yesp!

PR #69, opened 2020-01-09
Custom SpeechController Queue

kwylez:
This PR includes a better implementation of publishers and processing of remote audio files to be cached locally.

@noelweichbrodt


Spokestack/TextToSpeech.swift
```diff
@@ -209,26 +287,26 @@ import CryptoKit
     }
 }
 
-fileprivate struct TTSSSMLResponseURL: Codable {
-    let url: URL
+public struct TTSSSMLResponseURL: Codable {
```
noelweichbrodt: Why do these need to be public? Is it for the publisher?

Spokestack/TextToSpeech.swift
```diff
@@ -85,13 +90,86 @@ import CryptoKit
         self.synthesize(input: input, success: play)
     }
     
+    @available(iOS 13.0, *)
+    public func synthesizePublisher(_ input: TextToSpeechInput) -> AnyPublisher<TextToSpeechResult, Error> {
```
noelweichbrodt: Am I correct in thinking that `synthesizePublisher` is the single-input function, and `TextToSpeechQueue. synthesize` is the mutli-input function?

Spokestack/TextToSpeech.swift
```diff
@@ -37,6 +41,7 @@ import CryptoKit
     private lazy var player: AVPlayer = AVPlayer()
     private var apiKey: SymmetricKey?
     private let ttsInputVoices = [0: "demo-male"]
+    private let decoder = JSONDecoder()
```
noelweichbrodt: Is `decoder` as a class property to avoid allocating during every `synthesis` response?

Spokestack/TextToSpeech.swift
kwylez: I'm not sure why that was changed. that can be reverted

Spokestack/TextToSpeech.swift
kwylez: Correct

Spokestack/TextToSpeech.swift
kwylez: yes

Spokestack/TextToSpeech.swift
kwylez: Yes. was made `public` for the publisher. That will be reverted since the separate class will be deleted.

Spokestack/TextToSpeech.swift
```diff
@@ -92,47 +97,87 @@ import CryptoKit
...
+        }.eraseToAnyPublisher()
+        
+        return
+            createSynthesizeRequestFuture
+            .flatMap{urlRequst in
+                
+                URLSession.shared
+                .dataTaskPublisher(for: urlRequst)
+                .handleEvents(receiveSubscription: { _ in
+                  print("Network request will start")
```
noelweichbrodt: remove debug print please

PR #72, opened 2020-02-14
Defined NLU interface

noelweichbrodt:
Defines the client API for an NLU interface. Intended as a counterpart to https://github.com/spokestack/spokestack-android/pull/47.


Spokestack/NLUDelegate.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  NLUDelegate.swift
+//  Spokestack
+//
+//  Created by Noel Weichbrodt on 2/14/20.
+//  Copyright © 2020 Pylon AI, Inc. All rights reserved.
```
space-pope: Ping again on the copyright string in the new file template :)

Spokestack/NLUResult.swift
```diff
@@ -0,0 +1,64 @@
...
+        self.slots = slots
+    }
+}
+
+/// A slot extracted during intent classification.
+/// - Remark: Depending on the NLU service used, slots may be typed; if present, the type of each slot can be accessed with the `type` property.
+@objc public class Slot: NSObject {
+    
+    /// The underlying type of the slot value.
+    @objc public var type: String
```
space-pope: I had this discussion with @brentspell a bit ago—having the value as `Any?` allows the client to get the type via `type(of: slot.value)`, right? That would make this field redundant.

Spokestack/Tokenizer.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Spokestack
+//
+//  Created by Noel Weichbrodt on 2/14/20.
+//  Copyright © 2020 Pylon AI, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A simple protocol for Tokenizer services used by NLUService instances.
+@objc public protocol Tokenizer {
```
space-pope: Is the Wordpiece encoder meant to adopt this protocol? If so, I'd suggest a closer correspondence with the Android version (which I suppose I haven't released yet, so I was just expecting you to read my mind there...). It wraps tokenization and token encoding (into IDs for feeding the model) into a single step for efficiency, which has the additional effect of making it not a tokenizer but more of a token encoder.

In general, though, a tokenizer isn't a necessary piece of the top-level NLU infrastructure, as almost any third-party service we'd provide a plugin for would expect a string and do tokenization/token encoding internally. I'd hide this down alongside the TF NLU implementation since Spokestack isn't really a general-purpose NLP library.

Spokestack/NLUDelegate.swift
noelweichbrodt: Ah yes. Changed the template, I will find-replace the whole project after merge.

Spokestack/NLUResult.swift
noelweichbrodt: I believe that the dynamic value may not be an instance of the actual type due to serialization, so  `type` would be necessary in order to reconstitute `value` as the desired type. But then again i may just be assuming some type loss in theory that can't occur in practice.

PR #74, opened 2020-02-26
The Third Leg

noelweichbrodt:
Implements an on-device NLU based on a Tensorflow Bert model, with associated support and example code.


SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -0,0 +1,85 @@
...
+    func testParse() {
+        let config = SpeechConfiguration()
+        config.nluVocabularyPath = SharedTestMocks.createVocabularyPath()
+        let parser = try! NLUTensorflowSlotParser(configuration: config, inputMaxTokenLength: 100)
+        let metaData = createMetadata().data(using: .utf8)
+        let metadata = try! JSONDecoder().decode(NLUTensorflowMetadata.self, from: metaData!)
+        
+        // selset parse
+        let taggedInputLocation = zip(["b_location"],["kitchen"])
+        let parsedSelset = try! parser.parse(taggedInput: taggedInputLocation, intent: metadata.intents.filter({ $0.name == "request.lights.deactivate" }).first!)
```
space-pope: This file is sort of integration/component test-shaped in general (you're exercising JSON deserialization as well as the actual slot parsing)...but to help readability, these intents could be pulled out into their own variables.

SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -0,0 +1,85 @@
...
+        XCTAssertEqual(parsedSelset["location"]!.value as! String, "room")
+        
+        // integer parse
+        let taggedInputRating = zip(["b_rating"],["ten"])
+        let parsedInteger = try! parser.parse(taggedInput: taggedInputRating, intent: metadata.intents.filter({ $0.name == "rate.app" }).first!)
+        XCTAssertEqual(parsedInteger["rating"]!.value as! Int, 10)
+        
+        // digits parse
+        let taggedInputPhone = zip(["b_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number"],["42", "##38", "##34", "##17", "##45"])
```
space-pope: I think a test for the NLU component might be in order:
1) We're not exercising the code that would create the `taggedInput.*` data at runtime
2) Along those same lines, there's nothing that decodes a whole sequence (something that includes `o` tags and maybe more than one slot).

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -0,0 +1,73 @@
...
+//  Copyright © 2020 Pylon AI, Inc. All rights reserved.
+//
+
+import XCTest
+import Spokestack
+
+class NLUTensorflowTokenizerTest: XCTestCase {
+    
+    func testWordpieceTokenize() {
+        let tokenizer = WordpieceTokenizer(try! SharedTestMocks.createEncodingsDictionary(SharedTestMocks.createVocabularyPath()))
```
space-pope: There's a lot of temp file creation/reading going on in this test. You should be able to simplify it with a `setUp` for the class.

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -0,0 +1,73 @@
...
+        XCTAssertEqual(try tokenizer.detokenize(["there", "unto"]), "there unto")
+    }
+    
+    func testBertTokenize() {
+        let config = SpeechConfiguration()
+        config.nluVocabularyPath = SharedTestMocks.createVocabularyPath()
+        let tokenizer = try! BertTokenizer(config)
+        let text = "With her from — the one: this also has?"
+        let tokens = ["with", "her", "from", "—", "the", "one", ":", "this", "also", "has", "?"]
+        XCTAssertEqual(tokenizer.tokenize(text), tokens)
```
space-pope: Now detokenize it :)

Spokestack/Extensions/Data+Extensions.swift
```diff
@@ -9,12 +9,19 @@
...
         return withUnsafeBytes {
             Array(UnsafeBufferPointer(start: $0.bindMemory(to: T.self).baseAddress, count: count/MemoryLayout<T>.size))
         }
     }
+    
+    /// Convert Data into an array of type `T` with the specified number of elements.
+    /// - Parameters:
+    ///   - type: The type of the elements in the resulting array.
+    ///   - count: The number of elements in the resulting array.
+    func toArray<T>(type: T.Type, count: Int) -> [T] where T: ExpressibleByIntegerLiteral {
```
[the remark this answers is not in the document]
noelweichbrodt: https://i.kym-cdn.com/photos/images/newsfeed/001/443/076/68c.png

Spokestack/NLUTensorflow.swift
```diff
@@ -0,0 +1,186 @@
...
+    
+    /// Initializes an NLU instance.
+    /// - Note: An instance initialized this way is expected to use the pub/sub Combine interface, not the delegate interface, when calling `classify`.
+    /// - Requires: `SpeechConfiguration.nluVocabularyPath`, `SpeechConfiguration.nluTerminatorTokenIndex`, `SpeechConfiguration.nluPaddingTokenIndex`, `SpeechConfiguration.nluModelPath`, `SpeechConfiguration.nluModelMetadataPath`, and `SpeechConfiguration.nluMaxTokenLength`.
+    /// - Parameter configuration: Configuration parameters for the NLU.
+    @objc public init(configuration: SpeechConfiguration) throws {
+        self.configuration = configuration
+        self.terminatorToken = configuration.nluTerminatorTokenIndex
+        self.paddingToken = configuration.nluPaddingTokenIndex
+        super.init()
```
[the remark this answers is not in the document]
noelweichbrodt: swift insists that child classes init all properties before the explicit `super.init` call.

Spokestack/NLUTensorflow.swift
```diff
@@ -0,0 +1,186 @@
...
+            if intentsArgmax.0 > metadata.model.intents.count {
+                throw NLUError.model("NLU model returned an intent value outside the expected range.")
+            }
+            let intent = metadata.model.intents[intentsArgmax.0]
+            
+            // extract, decode + detokenize the classified tags, then hydrate the result slots based on the provided model metadata.
+            let encodedTagTensor = try model.output(at: OutputTensors.tag.rawValue)
+            let encodedTags = encodedTagTensor.data.toArray(type: Float32.self, count: encodedTagTensor.data.count/4)
+            // the posteriors for the tags are grouped by the number of model metadata tags, so stride through them calculating the argmax for each stride.
+            let encodedTagsArgmax = stride(from: 0,
```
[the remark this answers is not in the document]
noelweichbrodt: `stride` is a nice port from the C++ realm.

Spokestack/NLUTensorflowMetadata.swift
```diff
@@ -0,0 +1,62 @@
...
+    let slots: [NLUTensorflowSlot]
+}
+
+internal struct NLUTensorflowSlot: Codable {
+    let name: String
+    let type: String
+    let facets: String
+}
+
+internal struct NLUTensorflowSelset: Codable {
```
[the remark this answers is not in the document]
noelweichbrodt: Any change to the fascet type/representation/parsing is going to be a point release due to the overall coupling of the system no matter the extensibility of the slot fascet decoders.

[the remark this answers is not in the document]
space-pope: Sure; I was thinking more of the difficulty of adding new slot types (can an end user even do this?), not so much a change to the existing schemas.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -0,0 +1,169 @@
...
+            if let tokens = slotTokens[slotType] {
+                if let value = try tokenizer?.detokenize([token]) {
+                    slotTokens[slotType] = tokens + [value]
+                }
+            } else {
+                slotTokens[slotType] = [token]
+            }
+        }
+        // for each tag that isn't unclassified, send the tokens to the slot facet parser
+        for slotName in slotTokens.keys where slotName != "o" {
```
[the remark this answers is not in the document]
noelweichbrodt: It's the same op either way, but this way places it on the same line as the other relevant part of why it's done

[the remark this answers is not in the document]
space-pope: Not totally—removing the `o` key entirely means you don't have to do the equality check on each key you hit during the iteration over the map. It's admittedly a micro-optimization, but that's how you know it was really me who wrote the review.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -0,0 +1,176 @@
...
+            return piecewiseEncode(text: text, index: 1, pieces: [])
+        }
+    }
+    
+    /// Decode and detokenize the encoded tokens into a space-separated string.
+    /// - Parameter tokens: The encoded tokens to decode and detokenize.
+    func detokenize(_ tokens: [String]) throws -> String {
+        return tokens.reduce("", { (result, s) in
+            if s.prefix(2) == self.piecePrefix {
+                return result + s.suffix(s.count - 2)
```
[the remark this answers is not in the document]
noelweichbrodt: Yeah, i wasn't certain how stable the IOB tagging was going to be, but given what I now know it's fine to hardcode that in order to reduce the complexity here.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -0,0 +1,176 @@
...
+
+/// MARK: BertTokenizer
+
+/// Using the `BasicTokenizer` and `WordpieceTokenizer`, performs tokenization + encoding/detokenization + decoding specific to the BERT NLU model.
+internal struct BertTokenizer {
+    
+    /// The maximum input token count the BERT model supports.
+    internal var maxTokenLength: Int?
+    
+    private let minVocabLength = 2
```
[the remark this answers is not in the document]
noelweichbrodt: I should move this check outside of the `encode` function and into the initializer at least.

Spokestack/NLUTensorflowMetadata.swift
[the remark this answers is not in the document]
noelweichbrodt: In the android version, the slot parser interface would allow for a user-provided (eg one not included in the Spokestack package) slot parser to be configured?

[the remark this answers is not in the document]
space-pope: Yep; you just have to call `registerSlotParser` with a slot type and the name of your parser class (which has to implement `SlotParser`) when you're building the NLU object.

Spokestack/NLUTensorflow.swift
```diff
@@ -91,12 +91,16 @@ import TensorFlowLite
         self.interpreter = try Interpreter(modelPath: self.configuration.nluModelPath)
         try self.interpreter!.allocateTensors()
         if(self.interpreter!.inputTensorCount != InputTensors.allCases.count) || (self.interpreter!.outputTensorCount != OutputTensors.allCases.count) {
-            throw NLUError.model("NLU model provided is not shaped as expected. There are \(self.interpreter!.inputTensorCount)/\(InputTensors.allCases.count) inputs and \(self.interpreter!.outputTensorCount)/\(OutputTensors.allCases.count) outputs")
+            let inputCount = self.interpreter!.inputTensorCount
```
space-pope: These placeholders would do ever so slightly more good outside the check.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -25,8 +25,8 @@ internal struct BasicTokenizer: Tokenizer {
             })
             // convert the array of characters back to a string
             .joined()
-            //  split the string into alphanumeric word/punctuation components. Also normalize the string again, removing all but alphanumerics and punctuation.
-            .components(separatedBy: NSCharacterSet.alphanumerics.union(NSCharacterSet.punctuationCharacters).inverted)
+            //  split the string into alphanumeric (L* U N*) word/punctuation components. Also normalize the string again, removing all but alphanumerics and punctuation.
+            .components(separatedBy: NSCharacterSet.alphanumerics.subtracting(NSCharacterSet.nonBaseCharacters).union(NSCharacterSet.punctuationCharacters).inverted)
```
space-pope: Hm, OK. I think this works—makes things a bit harder to grok, but the character sets themselves seem right.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -153,24 +152,9 @@ internal struct BertTokenizer {
         try self.encode(self.tokenize(text))
     }
     
-    /// Decode the encoded tokens.
-    /// - Parameter encoded: The encoded tokens to decode.
-    func decode(_ encoded: [Int]) throws -> [String] {
```
space-pope: You can take out the `decodings` map and its population now too, right?

Spokestack/NLUTensorflow.swift
```diff
@@ -0,0 +1,186 @@
...
+            super.init()
+            try self.initializeInterpreter()
+            guard let model = self.interpreter else {
+                throw NLUError.model("NLU model was not initialized.")
+            }
+            let inputTensor = try model.input(at: InputTensors.input.rawValue)
+            let inputMaxTokenLength = inputTensor.shape.dimensions[1]
+            self.maxTokenLength = inputMaxTokenLength
+            self.tokenizer = try BertTokenizer(configuration)
+            self.tokenizer?.maxTokenLength = inputMaxTokenLength
```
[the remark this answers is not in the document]
noelweichbrodt: This is a good catch. I have threaded the max length through several components of the NLU inconsistently.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -0,0 +1,169 @@
...
+internal struct NLUTensorflowSlotParser {
+    private var tokenizer: BertTokenizer?
+    private let decoder = JSONDecoder()
+    
+    /// Initializes a slot parser for the specific nlu tensorflow model.
+    /// - Parameters:
+    ///   - configuration: The global SpeechConfiguration object.
+    ///   - inputMaxTokenLength: The maximum number of input tokens to the nlu tensorflow model.
+    init(configuration: SpeechConfiguration, inputMaxTokenLength: Int) throws {
+        self.tokenizer = try BertTokenizer(configuration)
```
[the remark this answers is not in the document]
noelweichbrodt: Another good find; in the context of the slot parser detokenization was just an identity operation and was a leftover from an earlier refactoring.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -0,0 +1,169 @@
...
+            guard let lowerBound = parsedSlot.range.first,
+                let upperBound = parsedSlot.range.last,
+                let parsedValue = parseNumber(text)
+                else {
+                return nil
+            }
+            let range = ClosedRange<Int>(uncheckedBounds: (lower: lowerBound, upper: upperBound))
+            return range.contains(parsedValue) ? parsedValue : nil
+        case "digits":
+            let parsedSlot = try decoder.decode(NLUTensorflowDigits.self, from: facetData)
```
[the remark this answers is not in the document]
noelweichbrodt: Not sure what this is trying to point out? The JSON deserialization is done once for each slot.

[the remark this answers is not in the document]
space-pope: Once for each slot, each time classification runs. What I was saying is that since the metadata is static, it would be best if its deserialization were a one-time cost rather than happening for each utterance that comes in.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -82,31 +82,33 @@ internal struct WordpieceTokenizer: Tokenizer {
         }
     }
     
-    /// Decode and detokenize the encoded tokens into a space-separated string.
-    /// - Parameter tokens: The encoded tokens to decode and detokenize.
-    func detokenize(_ tokens: [String]) throws -> String {
-        return tokens.reduce("", { (result, s) in
+    func decode(_ tokens: [String]) ->  [String] {
```
space-pope: This doesn't really decode (turn IDs into tokens); it's more the Wordpiece-specific half of detokenization.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -82,31 +82,33 @@ internal struct WordpieceTokenizer: Tokenizer {
...
-                return s
+                return [s]
             }
         })
     }
+    
+    /// Decode and detokenize the encoded tokens into a space-separated string.
+    /// - Parameter tokens: The encoded tokens to decode and detokenize.
+    func detokenize(_ tokens: [String]) -> String {
+        return self.decode(tokens).joined(separator: " ")
```
space-pope: The behavior is partially documented now, but the unconditional space separation and failure to retain the original text will still hurt display of the slot values.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -137,24 +137,25 @@ internal struct BertTokenizer {
...
     }
-    
+        
     /// Tokenize and encode the input text.
     /// - Parameter text: The input text to tokenize and encode.
     func tokenizeAndEncode(_ text: String) throws -> [Int] {
         try self.encode(self.tokenize(text))
     }
     
+    func decode(_ tokens: [String]) ->  [String] {
```
space-pope: Like the fn in the other tokenizer, not really decoding.

SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -11,28 +11,60 @@ import Spokestack
...
         
-        // integer parse
-        let taggedInputRating = zip(["b_rating"],["ten"])
-        let parsedInteger = try! parser.parse(taggedInput: taggedInputRating, intent: metadata.intents.filter({ $0.name == "rate.app" }).first!)
-        XCTAssertEqual(parsedInteger["rating"]!.value as! Int, 10)
+        let taggedInputRating51 = zip(["b_rating", "i_rating", "i_rating", "i_rating"],["fi","##ft", "##ie", "one"])
+        let parsedInteger51 = try! parser.parse(taggedInput: taggedInputRating51, intent: metadata!.intents.filter({ $0.name == "rate.app" }).first!, encoder: encoder!)
+        XCTAssertEqual(parsedInteger51["rating"]!.value as! Int, 51)
+
+        let taggedInputRatingNil = zip(["b_rating", "i_rating", "i_rating", "i_rating"],["fi","##ft", "##ie", "six"])
```
space-pope: I hope ASR doesn't give us "fiftie six" :)

SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -11,28 +11,60 @@ import Spokestack
...
+        let taggedInputPhoneNumeric = zip(["b_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number"],["42", "##38", "##34", "##17", "##45"])
+        let parsedDigitsNumeric = try! parser.parse(taggedInput: taggedInputPhoneNumeric, intent: metadata!.intents.filter({ $0.name == "inform.phone_number" }).first!, encoder: encoder!)
+        XCTAssertEqual(parsedDigitsNumeric["phone_number"]!.value as! String, "4238341745")
+
+        let taggedInputPhoneCardinal = zip(["b_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number", "i_phone_number"],["4", "second", "three", "eighth", "three", "four", "one", "seven", "four", "five"])
+        let parsedDigitsCardinal = try! parser.parse(taggedInput: taggedInputPhoneCardinal, intent: metadata!.intents.filter({ $0.name == "inform.phone_number" }).first!, encoder: encoder!)
+        XCTAssertEqual(parsedDigitsCardinal["phone_number"]!.value as! String, "4238341745")
+    }
+    
+    func testParseEntity() {
```
space-pope: It might be good to try an input or two with some `"o"` tags and more than one slot value just to exercise those parts of the parsing logic.

noelweichbrodt:
I suppose that it's too late to note that `BertTokenizer` changes are in progress 😬. Apologies for not mentioning that earlier!

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -39,99 +28,143 @@ internal struct NLUTensorflowSlotParser {
...
-                let upperBound = parsedSlot.range.last,
-                let parsedValue = parseNumber(text)
+            guard let parsed = try slot.parsed() as? NLUTensorflowInteger else {
+                throw NLUError.metadata("The NLU metadata for the \(slot.name) facet was not found.")
+            }
+            let integer = encoder
+                .decode(values)
+                .reduce([], { self.parseReduceNumber($0, next: $1) })
+                .reduce(0, { $0 + $1 })
+                //.compactMap({ parseNumber($0) })
```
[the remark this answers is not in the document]
noelweichbrodt: imo `parseReduceNumber` is already complicated enough, the second reduction is a much simpler expression than if the logic was moved into it.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -39,99 +28,143 @@ internal struct NLUTensorflowSlotParser {
...
+            guard let lowerBound = parsed.range.first,
+                let upperBound = parsed.range.last
                 else {
                 return nil
             }
             let range = ClosedRange<Int>(uncheckedBounds: (lower: lowerBound, upper: upperBound))
-            return range.contains(parsedValue) ? parsedValue : nil
+            return range.contains(integer) ? integer : nil
```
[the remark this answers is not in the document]
noelweichbrodt: the Swift syntax for doing multiple precedence operations on optional values as you suggest is just as complicated as the range operation :(

Spokestack/NLUTensorflowMetadata.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    /// Using the provided configuration that includes the `nluModelMetadataPath`, hydrate the corresponding the metadata structure.
+    /// - Parameter configuration: The global SpeechConfiguration.
+    init(_ configuration: SpeechConfiguration) throws {
+        self.configuration = configuration
+
+        let metadataURL = URL(fileURLWithPath: configuration.nluModelMetadataPath)
+        let metaData = try Data(contentsOf: metadataURL)
+        guard let metadata = try? JSONDecoder().decode(NLUTensorflowMetadata.self, from: metaData) else {
+            throw NLUError.metadata("Could not parse model metadata file set at nluModelMetadataPath.")
```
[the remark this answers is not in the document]
noelweichbrodt: The main `classify` function has two arities that deal with concurrency. The delegate arity performs all nlu work in a background thread, while the combine arity documents a warning to call it outside the main ui thread (which is the expected pattern for combine publishers, to defer to the consumer's decisions regarding threading).

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -39,99 +28,143 @@ internal struct NLUTensorflowSlotParser {
...
         }
         
         return slots
     }
     
-    private func slotFacetParser(slot: NLUTensorflowSlot, facetData: Data, values: [String]?) throws -> Any? {
-        guard let text = values else {
-            return nil
-        }
+    private func slotFacetParser(slot: NLUTensorflowSlot, values: [String], encoder: BertTokenizer) throws -> Any? {
```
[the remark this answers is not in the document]
noelweichbrodt: This is a good suggestion, but I'll defer this change to a later release.

noelweichbrodt:
This last set of commits should resolve all outstanding issues, in particular around reconstructing complicated slot values.

SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -26,45 +26,52 @@ class NLUTensorflowSlotParserTest: XCTestCase {
...
-        let parsedEntity = try! parser.parse(taggedInput: taggedInputEntity, intent: metadata!.intents.filter({ $0.name == "identify" }).first!, encoder: encoder!)
+        let taggedInputEntity = ["b_epynonymous", "i_epynonymous"]
+        let et1 = EncodedTokens(tokensByWhitespace: ["dead", "beef"], encodedTokensByWhitespaceIndex: [0, 1], encodedTokens: [0])
+        let parsedEntity = try! parser.parse(tags: taggedInputEntity, intent: metadata!.intents.filter({ $0.name == "identify" }).first!, encoder: encoder!, encodedTokens: et1)
         XCTAssertEqual(parsedEntity["epynonymous"]!.value as! String, "dead beef")
+        
+        let taggedInputEntityO = ["o", "b_epynonymous", "i_epynonymous", "o", "o", "b_epynonymous", "o"]
+        let etO = EncodedTokens(tokensByWhitespace: ["when", "dead", "beef", "appears", "in", "debug"], encodedTokensByWhitespaceIndex: [0, 1, 2, 3, 4, 5], encodedTokens: [0])
+        let parsedEntityO = try! parser.parse(tags: taggedInputEntityO, intent: metadata!.intents.filter({ $0.name == "identify" }).first!, encoder: encoder!, encodedTokens: etO)
+        XCTAssertEqual(parsedEntityO["epynonymous"]!.value as! String, "dead beef debug")
```
space-pope: Interesting. This would be a strange thing to get from the model. I think the Android version would just give you `"debug"` as the value here instead of concatenating the strings. We don't currently support or claim to support list/array values for slots.

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
-    func testBertTokenize() {
+
+    func testBertTokenizerEncode() {
         let config = SpeechConfiguration()
         config.nluVocabularyPath = SharedTestMocks.createVocabularyPath()
         let tokenizer = try! BertTokenizer(config)
-        let text = "With her from — the one: this also has?"
+        
+        let t2 = try! tokenizer.encode(text: "thereunto")
+        //XCTAssertEqual(t2.normalizedTokens, ["[UNK]"])
```
space-pope: Dead code in this test.

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        //XCTAssertEqual(try! tokenizer.encode(text: t4).normalizedTokens, tokens)
         
         let phone = "4238341745"
         let phoneTokens = ["42", "##38", "##34", "##17", "##45"]
-        XCTAssertEqual(tokenizer.tokenize(phone), phoneTokens)
+        //XCTAssertEqual(try! tokenizer.encode(text: phone).normalizedTokens, phoneTokens)
+    
+        let text = "With her from—the one: this also has?"
+        let encoded = [18, 25, 24, 1, 7, 39, 4, 34, 47, 49, 5]
+        let indicies = [0, 1, 2, 2, 2, 3, 3, 4, 5, 6, 6]
```
space-pope:
```suggestion
        let indices = [0, 1, 2, 2, 2, 3, 3, 4, 5, 6, 6]
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        let wt = ["With", "her", "from—the", "one:", "this", "also", "has?"]
+        let et = try! tokenizer.encode(text: text)
+        XCTAssertEqual(et.encodedTokens, encoded)
+        XCTAssertEqual(et.encodedTokensByWhitespaceIndex, indicies)
```
space-pope:
```suggestion
        XCTAssertEqual(et.encodedTokensByWhitespaceIndex, indices)
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
         let tokenizer = try! BertTokenizer(config)
-        let text = "this also has"
-        XCTAssertEqual(try tokenizer.detokenize(tokenizer.tokenize(text)), text)
+        
+        let t1 = "Godspeed You! Black Emperor"
+        let tw1 = ["Godspeed", "You!", "Black", "Emperor"]
+        let etw1 = [0]
+        let etet1 = [0]
+        let et1 = EncodedTokens(tokensByWhitespace: tw1, encodedTokensByWhitespaceIndex: etw1, encodedTokens: etet1)
+        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et1, whitespaceIndicies: Array(0...3)), t1)
```
space-pope:
```suggestion
        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et1, whitespaceIndices: Array(0...3)), t1)
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+    func testBertTokenizerdecodeWithWhitespace() {
         let config = SpeechConfiguration()
         config.nluVocabularyPath = SharedTestMocks.createVocabularyPath()
         let tokenizer = try! BertTokenizer(config)
-        let text = "this also has"
-        XCTAssertEqual(try tokenizer.detokenize(tokenizer.tokenize(text)), text)
+        
+        let t1 = "Godspeed You! Black Emperor"
+        let tw1 = ["Godspeed", "You!", "Black", "Emperor"]
+        let etw1 = [0]
```
space-pope: Again, why is there only one encoded ID? I guess it's not relevant to the current test; it's just hard to tell what's going on.

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        let etet1 = [0]
+        let et1 = EncodedTokens(tokensByWhitespace: tw1, encodedTokensByWhitespaceIndex: etw1, encodedTokens: etet1)
+        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et1, whitespaceIndicies: Array(0...3)), t1)
+
+        var et2 = EncodedTokens()
+        let t2 = "their"
+        et2.tokensByWhitespace = [t2]
+        et2.encodedTokensByWhitespaceIndex = [0]
+        et2.encodedTokens = [48]
+        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et2, whitespaceIndicies: [0]), t2)
```
space-pope:
```suggestion
        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et2, whitespaceIndices: [0]), t2)
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        
+        var et3 = EncodedTokens()
+        let t3 = "theres"
+        et3.tokensByWhitespace = [t3]
+        et3.encodedTokensByWhitespaceIndex = [0,0]
+        et3.encodedTokens = [56, 26]
+        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et3, whitespaceIndicies: [0]), t3)
```
space-pope:
```suggestion
        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: et3, whitespaceIndices: [0]), t3)
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
     }
     
-    func testBertTokenizerEncode() {
+    func testBertEncodeDecodeRoundtrip() {
         let config = SpeechConfiguration()
         config.nluVocabularyPath = SharedTestMocks.createVocabularyPath()
         let tokenizer = try! BertTokenizer(config)
+        let t1 = "this also has"
+        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: try! tokenizer.encode(text: t1), whitespaceIndicies: Array(0...2)), t1)
```
space-pope:
```suggestion
        XCTAssertEqual(try! tokenizer.decodeWithWhitespace(encodedTokens: try! tokenizer.encode(text: t1), whitespaceIndices: Array(0...2)), t1)
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        
         let text = "With her from—the one: this also has?"
-        let encoded = [18, 25, 24, 1, 7, 39, 4, 34, 47, 49, 5]
-        XCTAssertEqual(try! tokenizer.encode(tokenizer.tokenize(text)), encoded)
+        let et = try! tokenizer.encode(text: text)
+        let decodedFull = try! tokenizer.decodeWithWhitespace(encodedTokens: et, whitespaceIndicies: Array(0...et.tokensByWhitespace!.count-1))
```
space-pope:
```suggestion
        let decodedFull = try! tokenizer.decodeWithWhitespace(encodedTokens: et, whitespaceIndices: Array(0...et.tokensByWhitespace!.count-1))
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        XCTAssertEqual(decodedFull, text)
+        let decodedPartial = try! tokenizer.decodeWithWhitespace(encodedTokens: et, whitespaceIndicies: [1,2])
```
space-pope:
```suggestion
        let decodedPartial = try! tokenizer.decodeWithWhitespace(encodedTokens: et, whitespaceIndices: [1,2])
```

SpokestackTests/NLUTensorflowTokenizerTest.swift
```diff
@@ -10,56 +10,75 @@ import XCTest
...
+        XCTAssertEqual(decodedPartial, "her from—the")
```
space-pope: I'd sort of like to see a test that roundtrips a string containing a codepoint or two that gets normalized away (say, a diacritic), but meh.

Spokestack/NLUResult.swift
```diff
@@ -26,6 +26,10 @@ import Foundation
     /// The slot values present in the utterance.
     @objc public var slots: [String:Slot]
     
+    public override var description: String {
+        return "\(intent) \(confidence) \(slots)"
```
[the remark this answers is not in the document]
noelweichbrodt: the slot uses it's `description` property when called in this manner. the full output is `request.lights.activate 0.9998938 ["location": (selset: room)]`, which is think is very clear.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -14,77 +14,79 @@ internal struct NLUTensorflowSlotParser {
...
             // the model slot recognizer uses IOB tags, so `b_` and `i_` prefixes must be removed to resolve tag labels to slot names.
             var slotType = tag
+            let whitespaceIndex = encodedTokensByWhitespaceIndex[index]
             if let prefixIndex = tag.range(of: "_")?.upperBound {
                 slotType = String(tag.suffix(from: prefixIndex))
             }
             // collect all the tokens with the same slot type into a single array
-            if let tokens = slotTokens[slotType] {
-                slotTokens[slotType] = tokens + [token]
+            if slotDictionary.keys.contains(where: { $0 == slotType }) {
```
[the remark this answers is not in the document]
noelweichbrodt: big oof :)

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -14,77 +14,79 @@ internal struct NLUTensorflowSlotParser {
...
     /// - Note: This operation effectively truncates the tag labels by the input size, ignoring all labels outside the input token count.
     /// - Parameters:
-    ///   - taggedInput: A zip of tokenized inputs and classification tag labels.
+    ///   - tags: An array of classification tag labels.
     ///   - intent: The predicted intent.
-    internal func parse(taggedInput: Zip2Sequence<[String], [String]>, intent: NLUTensorflowIntent, encoder: BertTokenizer) throws -> [String:Slot] {
+    internal func parse(tags: [String], intent: NLUTensorflowIntent, encoder: BertTokenizer, encodedTokens: EncodedTokens) throws -> [String:Slot] {
         
+        guard let encodedTokensByWhitespaceIndex = encodedTokens.encodedTokensByWhitespaceIndex else {
+            throw NLUError.tokenizer("NLU model tokenizer failed to correctly encode utterance because encodedTokensByWhitespaceIndex is nil.")
```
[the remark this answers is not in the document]
noelweichbrodt: It's good form to guard inside of each function that uses an Optional instead of assuming implicit dependencies, but this error message is definitely copy-pasta.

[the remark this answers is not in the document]
space-pope: Yeah...what I meant was that I don't think that field should be an Optional. If you don't have that information, something went wrong during tokenization, and classification shouldn't be attempted.

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -14,77 +14,79 @@ internal struct NLUTensorflowSlotParser {
...
+        }
+
         var slots: [String:Slot] = [:]
         
         // create a dictionary of [tags: [tokens]]
-        var slotTokens: [String:[String]] = [:]
-        for (tag, token) in taggedInput {
+        var slotDictionary: [String: [Int]] = [:]
+        for (index, tag) in tags.enumerated()  where index < encodedTokensByWhitespaceIndex.count{
```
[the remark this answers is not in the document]
noelweichbrodt: the zip idea lead me down a good refactoring path, thanks.

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -8,14 +8,82 @@
...
+    ///   - encodedTokens: The tokens to decode.
+    ///   - whitespaceIndicies: The desired indicies from `encodedTokensByWhitespaceIndex`.
+    func decode(encodedTokens: EncodedTokens, whitespaceIndicies: [Int]) throws -> [String] {
+        guard let tokensByWhitespace = encodedTokens.tokensByWhitespace else {
+            throw NLUError.tokenizer("NLU model tokenizer did not encoded tokens for this range.")
+        }
+        // only unique numbers in the index will be reconstructed. This prevents duplications arising from wordpiece labeling.
+        var uniques: Set<Int> = []
+        return whitespaceIndicies
+            // Filtering a set is O(1) vs O(n*2) reduce.
```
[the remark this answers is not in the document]
noelweichbrodt: `reduce` is better here, yeah

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -29,133 +97,45 @@ internal struct BasicTokenizer: Tokenizer {
...
-        return wordpieceTokenizer.detokenize(tokens)
-    }
+/// A simple data structure for tokenization and ecnoding an utterance, and reconstructing it.
+internal struct EncodedTokens {
+    /// An uterrance, tokenized by whitespace.
+    public var tokensByWhitespace: [String]?
+    /// Each encoded token entry is represented by an index back to the whitespaced token.
+    public var encodedTokensByWhitespaceIndex: [Int]?
+    /// Whitespaced tokens, tokenized, wordpieced, and encoded for input into BERT.
+    public var encodedTokens: [Int]?
```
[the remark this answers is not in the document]
noelweichbrodt: …and doesn't have to be optional anymore!

SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -26,45 +26,52 @@ class NLUTensorflowSlotParserTest: XCTestCase {
...
     func testParseInteger() {
-        let taggedInputRating10 = zip(["b_rating"],["ten"])
-        let parsedInteger10 = try! parser.parse(taggedInput: taggedInputRating10, intent: metadata!.intents.filter({ $0.name == "rate.app" }).first!, encoder: encoder!)
+        let et1 = EncodedTokens(tokensByWhitespace: ["ten"], encodedTokensByWhitespaceIndex: [0], encodedTokens: [0])
+        let parsedInteger10 = try! parser.parse(tags: ["b_rating"], intent: metadata!.intents.filter({ $0.name == "rate.app" }).first!, encoder: encoder!, encodedTokens: et1)
         XCTAssertEqual(parsedInteger10["rating"]!.value as! Int, 10)
         
-        let taggedInputRating51 = zip(["b_rating", "i_rating", "i_rating", "i_rating"],["fi","##ft", "##ie", "one"])
-        let parsedInteger51 = try! parser.parse(taggedInput: taggedInputRating51, intent: metadata!.intents.filter({ $0.name == "rate.app" }).first!, encoder: encoder!)
+        let et2 = EncodedTokens(tokensByWhitespace: ["fiftie", "one"],  encodedTokensByWhitespaceIndex: [0, 0, 0, 1], encodedTokens: [0])
```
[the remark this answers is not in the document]
noelweichbrodt: `encodedTokens` isn't used for slot parsing, just for model input. i'll add a comment to that effect.

SpokestackTests/NLUTensorflowSlotParserTest.swift
noelweichbrodt: oh interesting, i thought that entities constituted all of a single tag, not just a range.

noelweichbrodt: i…don't think the discrepancy is problematic, but defer to you on that.

SpokestackTests/NLUTensorflowTokenizerTest.swift
noelweichbrodt: Allow me to introduce you to the beautiful city of Ålesund!

Spokestack/NLUTensorflowSlotParsers.swift
```diff
@@ -16,41 +16,30 @@ internal struct NLUTensorflowSlotParser {
...
+                }
+            })
+            // send each tag's tokens through the slot parser and return the result
+            .reduce(into: [:] as [String:Slot], { result, dictionaryEntry in
+                let (tag, whitespaceIndices) = dictionaryEntry
+                guard let slot = intent.slots.filter({ $0.name == tag }).first else {
+                    throw NLUError.metadata("Could not find a slot called \(tag) in nlu model metadata.")
+                }
+                let slotValue = try self.slotFacetParser(slot: slot, whitespaceIndices: whitespaceIndices, encoder: encoder, encodedTokens: encodedTokens)
+                result.updateValue(Slot(type: slot.type, value: slotValue), forKey: tag)
```
space-pope: You start with an empty dict, so isn't this just `result[tag] = Slot(type: slot.type, value: slotValue)` since the incoming keys are already unique?

Spokestack/NLUTensorflowTokenizer.swift
```diff
@@ -38,23 +38,25 @@ internal struct BertTokenizer {
...
             normalizedTokens += tokens
             // finally encode the normalized tokens
             for t in tokens {
-                encoded.append(self.encodings[t] ?? -1)
+                if let code = self.encodings[t] {
+                    encoded.append(code)
+                } else if let unk = self.encodings[unknownPiece] {
+                    encoded.append(unk)
+                } else {
+                    throw NLUError.tokenizer("NLU tokenizer could not encode the tokenized text.")
```
space-pope: Not having an encoding for `[UNK]` is really due to a bad vocab file, and that's probably better checked for at initialization time rather than the first time you happen to hit one in input text.

SpokestackTests/SharedTestMocks.swift
```diff
@@ -94,6 +94,8 @@ internal struct SharedTestMocks {
     ##38
     ##45
     million
+    [UNK]
+    Ålesund
```
space-pope: There shouldn't actually be a diacritic in the vocab file; what you're testing is lossless reconstruction of an input that wasn't in the BERT vocab.

PR #77, opened 2020-04-02
SpeechMarkdown support for TTS input

noelweichbrodt:
`TextToSpeech` support for [Speech Markdown](https://www.speechmarkdown.org/) via Spokestack API's `synthesizeMarkdown` method.

### DIFF
--- a/Spokestack/TextToSpeech.swift
+++ b/Spokestack/TextToSpeech.swift
@@ -336,13 +336,13 @@ private let apiQueue = DispatchQueue(label: TTSSpeechQueueName, qos: .userInitia
 
 // MARK: Internal data structures
 
-// SSML
-fileprivate struct TTSSSMLResponseURL: Codable {
+fileprivate struct TTSResponseURL: Codable {
     let url: URL
 }
 
+// SSML
 fileprivate struct TTSSSMLResponseSynthesize: Codable {
-    let synthesizeSsml: TTSSSMLResponseURL
+    let synthesizeSsml: TTSResponseURL
 }
 
 fileprivate struct TTSSSMLResponseData: Codable {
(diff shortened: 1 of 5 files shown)


Spokestack/TextToSpeech.swift
```diff
@@ -335,6 +349,20 @@ fileprivate struct TTSSSMLResponseData: Codable {
     let data: TTSSSMLResponseSynthesize
 }
 
+// Markdown
+fileprivate struct TTSMarkdownResponseURL: Codable {
```
space-pope, line 353: Couldn't you have one `TTSResponseURL` to rule them all since the field is named the same across the board (and should never change)?

SpokestackTests/TextToSpeechTest.swift
```diff
@@ -51,6 +56,22 @@ class TextToSpeechTest: XCTestCase {
...
+    func testSynthesizeMarkdown() {
+        let delegate = TestTextToSpeechDelegate()
+        let config = SpeechConfiguration()
+        let tts = TextToSpeech(delegate, configuration: config)
+        
+        // successful request with markdown formatting
+        delegate.reset()
+        let didSucceedExpectation3 = expectation(description: "successful request calls TestTextToSpeechDelegate.success")
+        delegate.asyncExpectation = didSucceedExpectation3
+        let markdownInput = TextToSpeechInput("Yet right now the average age of this (52nd)[number] Parliament is (49)[number] years old, [1s] OK Boomer.</speak>", voice: .demoMale, inputFormat: .markdown)
```
space-pope, line 69: I went on a dark journey to see whether `(52nd)[number]` would work. Turns out it will, but it's not really the right way to express this in SMD. Spokestack doesn't totally support the "right" way, though, so carry on.

...however, the trailing `</speak>` is definitely wrong, and should either cause an error or some unexpected results if passed through the production synthesizer.

PR #79, opened 2020-04-15
fixed bug in nlu slot selset with pre/post-pended punctuation

noelweichbrodt:
An nlu sentence like `Turn the lights on in the kitchen.` would not result in a slot value for `kitchen` due to the `.`. This fixes that bug.

### DIFF
--- a/SpokestackTests/NLUTensorflowSlotParserTest.swift
+++ b/SpokestackTests/NLUTensorflowSlotParserTest.swift
@@ -29,6 +29,11 @@ class NLUTensorflowSlotParserTest: XCTestCase {
         let et = EncodedTokens(tokensByWhitespace: ["kitchen"], encodedTokensByWhitespaceIndex: [0], encoded: [0])
         let parsedSelset = try! parser.parse(tags: ["b_location"], intent: metadata!.intents.filter({ $0.name == "request.lights.deactivate" }).first!, encoder: encoder!, encodedTokens: et)
         XCTAssertEqual(parsedSelset!["location"]!.value as! String, "room")
+        
+        // selset value has appended punctuation
+        let et2 = EncodedTokens(tokensByWhitespace: ["kitchen."], encodedTokensByWhitespaceIndex: [0], encoded: [0,1])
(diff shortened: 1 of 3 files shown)


SpokestackTests/NLUTensorflowSlotParserTest.swift
```diff
@@ -29,6 +29,11 @@ class NLUTensorflowSlotParserTest: XCTestCase {
         let et = EncodedTokens(tokensByWhitespace: ["kitchen"], encodedTokensByWhitespaceIndex: [0], encoded: [0])
         let parsedSelset = try! parser.parse(tags: ["b_location"], intent: metadata!.intents.filter({ $0.name == "request.lights.deactivate" }).first!, encoder: encoder!, encodedTokens: et)
         XCTAssertEqual(parsedSelset!["location"]!.value as! String, "room")
+        
+        // selset value has appended punctuation
+        let et2 = EncodedTokens(tokensByWhitespace: ["kitchen."], encodedTokensByWhitespaceIndex: [0], encoded: [0,1])
+        let parsedSelset2 = try! parser.parse(tags: ["b_location"], intent: metadata!.intents.filter({ $0.name == "request.lights.deactivate" }).first!, encoder: encoder!, encodedTokens: et2)
```
space-pope, line 35: The test setup is throwing me here again. I don't know whether it's just that some of these fields don't actually matter, or if this is actually the expected tag sequence for this input. If it's the latter, something else is wrong—the punctuation mark should be its own token and should have an `"o"` tag.

noelweichbrodt: Yeah, test setup is unclear. The  `encodedTokensByWhitespaceIndex` and `encoded` do not matter past the first value in each.

space-pope: Yeah, I remembered something like that being the case for `EncodedTokens`; it was mostly the value of `tags` that I wanted to check on here.

noelweichbrodt: Oh right. The test setup is making some simplifications to the input that would not occur in practice because of the previous steps in the encoding/parsing pipeline. Although the `tokensByWhitespace` is accurate since that is the whitespace-separated utterance, but the rest of the tags don't matter since the test of focused exclusively on the `location` slot.

PR #85, opened 2020-07-15
Refactored SpeechPipeline with stages and profiles

noelweichbrodt:
A complete minor rewrite of the `SpeechPipeline` and associated components. This rewrite has three goals:

1. Testability for previously under-tested components
2. The pipeline utilizes stages that are completely configurable
3. The pipeline may be instantiated from a set of profiles that encapsulate configuration complexity

The end result is a more flexible, better-tested pipeline that better congrues with that in `spokestack-android`.


SpokestackTests/AudioControllerTest.swift
```diff
@@ -14,59 +14,117 @@ import AVFoundation
...
-        let setupFailedExpectation = expectation(description: "testStartStreaming calls AudioControllerTestDelegate as the result of setupFailed method completion")
-        let processFrameExpectation = expectation(description: "testStartStreaming calls AudioControllerTestDelegate as the result of processFrame method completion")
+        controller.configuration = config
+        controller.context = context
+        let delegate = AudioControllerTestDelegate(config, context: context)
+        context.listeners = [delegate]
+        context.stageInstances = [delegate]
+        let setupFailedExpectation = expectation(description: "testStartStreaming calls AudioControllerTestDelegate as the result of failure method completion")
 
         // Uninititalized delegates do not cause an exception during startStreaming
```
space-pope, line 26:
```suggestion
        // Uninitialized delegates do not cause an exception during startStreaming
```

SpokestackTests/AudioControllerTest.swift
```diff
@@ -14,59 +14,117 @@ import AVFoundation
...
+    }
+    
+    func testProcess() {
+        let controller = AudioController.sharedInstance
+        let config = SpeechConfiguration()
+        let context = SpeechContext()
+        controller.configuration = config
+        controller.context = context
+        let delegate = AudioControllerTestDelegate(config, context: context)
+        controller.context?.listeners = [delegate]
```
space-pope, line 57:
```suggestion
        context.listeners = [delegate]
```

SpokestackTests/AudioControllerTest.swift
```diff
@@ -14,59 +14,117 @@ import AVFoundation
...
+        context.stageInstances = [delegate]
+        let processFrameExpectation = expectation(description: "testStartStreaming calls AudioControllerTestDelegate as the result of processFrame method completion")
 
         // AudioControllerDelegate processFrame is called
         XCTAssertNoThrow(try AVAudioSession.sharedInstance().setCategory(.record))
         delegate.asyncExpectation = processFrameExpectation
-        XCTAssertNotNil(controller.delegate)
         XCTAssertNotNil(delegate.asyncExpectation)
```
space-pope, line 64: Does this line test anything in your code?

SpokestackTests/AudioControllerTest.swift
```diff
@@ -14,59 +14,117 @@ import AVFoundation
...
 }
 
 /// Spy pattern for the system under test.
-class AudioControllerTestDelegate: AudioControllerDelegate, PipelineDelegate {
+class AudioControllerTestDelegate: SpeechProcessor, SpeechEventListener {
+    var configuration: SpeechConfiguration
     
+    var context: SpeechContext
     var didProcessFrame: Bool = false
     var didSetupFailed: Bool = false
```
space-pope, line 81: `didSetupFail` or `setupFailed`, maybe?

SpokestackTests/AudioControllerTest.swift
```diff
@@ -14,59 +14,117 @@ import AVFoundation
...
         didSetupFailed = false
+        didProcessFrame = false
         asyncExpectation = .none
     }
     
+    func startStreaming() {}
+    
+    func stopStreaming() {}
+    
+    func didActivate() { }
```
space-pope, line 100:
```suggestion
    func didActivate() {}
```

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
-        delegate = SpeechPipelineTestDelegate()
-        delegate.asyncExpectation = activateExpectation
-        p.setDelegates(delegate, pipelineDelegate: delegate)
-        
-        // assert that the pipeline's delegate references (and pipeline's service delegates) have changed
-        XCTAssert(delegate === p.speechDelegate)
-        p.speechDelegate?.didActivate()
-        wait(for: [activateExpectation], timeout: 1)
-        XCTAssert(delegate.activated)
+        XCTAssert(p.configuration.fftHopLength == 30)
```
space-pope, line 45: This looks like a job for `XCTAssertEqual`.

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
     }
     
     /// activate & deactivate
     func testActivateDeactivate() {
         let didInitExpectation = expectation(description: "didInitExpectation fulfills when testActivateDeactivate calls SpeechPipelineTestDelegate as the result of didInit method completion")
+        let didActivateExpectation = expectation(description: "didActivateExpectation fulfills when testActivateDeactivate calls SpeechPipelineTestDelegate as the result of activate method completion")
+        let didDeActivateExpectation = expectation(description: "didDeActivateExpectation fulfills when testActivateDeactivate calls SpeechPipelineTestDelegate as the result of deactive method completion")
```
space-pope, line 52:
```suggestion
        let didDeactivateExpectation = expectation(description: "didDeactivateExpectation fulfills when testActivateDeactivate calls SpeechPipelineTestDelegate as the result of deactivate method completion")
```

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
         let delegate = SpeechPipelineTestDelegate()
-        delegate.asyncExpectation = didInitExpectation
         let config = SpeechConfiguration()
+        let context = SpeechContext()
 
         // init the pipeline
-        let p = SpeechPipeline(TestProcessor(true), speechConfiguration: config, speechDelegate: delegate, wakewordService: TestProcessor(), pipelineDelegate: delegate)
+        delegate.asyncExpectation = didInitExpectation
+        let p = SpeechPipeline(configuration: config, listeners: [delegate])
+        _ = TestProcessor(true, config: config, context: context)
```
space-pope, line 60: why is this initialized and thrown away?

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
         wait(for: [didInitExpectation], timeout: 1)
         
         // activate and deactivate the pipeline
+        delegate.reset()
+        delegate.asyncExpectation = didActivateExpectation
         p.activate()
+        wait(for: [didActivateExpectation], timeout: 1)
         XCTAssert(p.context.isActive)
+        delegate.deactivateExpectation = didDeActivateExpectation
```
space-pope, line 69:
```suggestion
        delegate.deactivateExpectation = didDeactivateExpectation
```

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
         p.deactivate()
+        wait(for: [didDeActivateExpectation], timeout: 1)
```
space-pope, line 71:
```suggestion
        wait(for: [didDeactivateExpectation], timeout: 1)
```

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
+        // tflite
+        delegate.asyncExpectation = didInit1Expectation
+        let expectedTFStages: [SpeechProcessors] = [.vad, .tfLiteWakeword, .appleSpeech]
+        let p1 = SpeechPipelineBuilder()
+            .useProfile(.tfLiteWakewordAppleSpeech)
+            .setListener(delegate)
+            .build()
+        wait(for: [didInit1Expectation], timeout: 1)
+        XCTAssertEqual(expectedTFStages, p1.configuration.stages)
+        XCTAssert(delegate === p1.context.listeners[0])
```
space-pope, line 161: `XCTAssertEqual` (here and below), or doesn't that do instance equality?

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -30,142 +31,195 @@ class SpeechPipelineTest: XCTestCase {
...
+        
+        // appleWW
+        delegate.reset()
+        let expectedAppleWWStages: [SpeechProcessors] = [.vad, .appleWakeword, .appleSpeech]
+        let wakeActiveMax = 10000
+        let p2 = SpeechPipelineBuilder()
+            .useProfile(.appleWakewordAppleSpeech)
+            .setProperty("wakeActiveMax", wakeActiveMax.description)
+            .build()
+        //wait(for: [didInit2Expectation], timeout: 1)
```
space-pope, line 171: dead

SpokestackTests/TFLiteWakewordRecognizerTest.swift
```diff
@@ -14,11 +14,11 @@ import XCTest
 
 class TFLiteWakewordRecognizerTest: XCTestCase {
     
-    let tflwr = TFLiteWakewordRecognizer.sharedInstance
     let context = SpeechContext()
     let delegate = TFLiteWakewordRecognizerTestDelegate()
     let config = SpeechConfiguration()
-    
+    var tflwr: TFLiteWakewordRecognizer?
```
space-pope, line 20: Why bother making it optional in tests?

SpokestackTests/TFLiteWakewordRecognizerTest.swift
```diff
@@ -153,21 +163,23 @@ fileprivate enum MockWakewordModels {
     }()
 }
 
-class TFLiteWakewordRecognizerTestDelegate: PipelineDelegate, SpeechEventListener {
+class TFLiteWakewordRecognizerTestDelegate: SpeechEventListener {
     // Spy pattern for the system under test.
     // asyncExpectation lets the caller's test know when the delegate has been called.
     var didError: Bool = false
     var deactivated: Bool = false
```
space-pope, line 170: These would be good names for `didDidActivate` et al in one of the other tests.

SpokestackTests/TFLiteWakewordRecognizerTest.swift
```diff
@@ -195,6 +207,12 @@ class TFLiteWakewordRecognizerTestDelegate: PipelineDelegate, SpeechEventListene
     
     func didDeactivate() {
         self.deactivated = true
+        guard let _ = self.didDeactivateExpectation else {
```
space-pope, line 210: or `if self.didDeactivateExpectation == nil {`

Spokestack/SpeechPipeline.swift
```diff
@@ -7,184 +7,195 @@
...
+    /// The pipeline starts in a deactivated state, awaiting either a triggered activation  from a wakeword or VAD, or an explicit call to `activate`.
     @objc public func start() -> Void {
-        if (self.context.isActive) {
-            self.stop()
-        }
-        AudioController.sharedInstance.startStreaming(context: self.context)
-        self.wakewordRecognizerService.startStreaming(context: self.context)
-        self.speechConfiguration?.delegateDispatchQueue.async {
-            self.pipelineDelegate?.didStart()
+
```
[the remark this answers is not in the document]
space-pope: Update: Android just got an adjustment in this area. You'll still want to avoid the list of stages growing without bound, but calling start on a running pipeline isn't necessarily an error condition. See https://github.com/spokestack/spokestack-android/pull/81

noelweichbrodt:
As always, thank you for the *thorough* review. I'll address individual points as I go through, but to answer your two main points:

> Delegating full control of the processing loop to the individual stages
This was definitely an intermediate step design decision. I do want to move control fully over to the pipeline, but this decision removed complexity from some of the other refactors that were unavoidable. 

> user extensibility
User extensibility was not a priority for this refactor (thought the work is intended to set that up for later) since that use case is not currently required by any clients. I do like your suggestion of refactoring the profiles into concrete classes, will make that future work.

Spokestack/AudioController.swift
```diff
@@ -69,14 +71,10 @@ class AudioController {
...
-    public weak var pipelineDelegate: PipelineDelegate?
     /// Configuration for the audio controller.
-    public var configuration: SpeechConfiguration = SpeechConfiguration()
-
+    public var configuration: SpeechConfiguration?
+    public var context: SpeechContext?
+    
     // MARK: Private (properties)
     
     // private var bufferDuration: TimeInterval = TimeInterval((configuration.sampleRate / 1000) * configuration.frameWidth)
```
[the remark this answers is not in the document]
noelweichbrodt: yep. able to shrink the audiocontroller surface area with this refactor

Spokestack/Error.swift
```diff
@@ -13,6 +13,8 @@ import Foundation
 public enum AudioError: Error, Equatable {
     /// An audio unit system error
     case audioSessionSetup(String)
+    /// An audio controller
```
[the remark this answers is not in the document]
noelweichbrodt: the comments are solely for the autogen docs, but these comments are yeah, a little redundant.

Spokestack/SpeechContext.swift
```diff
@@ -14,10 +14,12 @@ import Foundation
     @objc public var transcript: String = ""
     /// Current speech recognition confidence: [0-1)
     @objc public var confidence: Float = 0.0
-    /// Speech pipeline active indicator
-    @objc public var isStarted: Bool = false
     /// Speech recognition active indicator
     @objc public var isActive: Bool = false
-    /// Speech detected indicator
-    @objc public var isSpeech: Bool = false
+    /// Speech detected indicator. Default to true for non-vad activation
```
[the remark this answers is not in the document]
noelweichbrodt: this is used as part of the pipeline for vad and vad-listening components to adjust their state off of.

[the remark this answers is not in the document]
space-pope: Re-reading this, I should've been more specific. In case it wasn't clear, it was the default of `true` that I was objecting to, since I don't think `isSpeech` should be true unless there's a VAD telling you that speech has happened, or something's manually activated the pipeline.

Spokestack/SpeechPipeline.swift
```diff
@@ -7,184 +7,195 @@
...
+                    return VADTrigger(self.configuration, context: self.context)
+                }
+            }()
+            self.context.stageInstances.append(stageInstance)
+        })
+        
+        // notify stages to start
+        AudioController.sharedInstance.startStreaming()
+        self.context.stageInstances.forEach { stage in
+            stage.startStreaming()
```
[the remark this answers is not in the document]
noelweichbrodt: stages require differing thread QoS, so it's left to the stage to determine threading.

Spokestack/SpeechPipeline.swift
```diff
@@ -7,184 +7,195 @@
...
+            listener.didStop()
+        })
+        self.context.stageInstances = []
+    }
+}
+
+/**
+    Convenince initializer for building a `SpeechPipeline` instance using a pre-configured profile. A pipeline profile encapsulates a series of configuration values tuned for a specific task.
+ 
+    Profiles are not authoritative; they act just like calling a series of methods on a {@link SpeechPipeline.Builder}, and any configuration properties they set can be overridden by subsequent calls.
```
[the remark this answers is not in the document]
noelweichbrodt: mmm copypasta

Spokestack/SpeechProcessor.swift
```diff
@@ -14,21 +14,61 @@ import Foundation
...
+    case tfLiteWakewordAppleSpeech
+    /// VAD-triggered Apple ASR
+    case vadTriggerAppleSpeech
+    /// Apple ASR that is manually activated and deactivated
+    case pushToTalkAppleSpeech
+}
+
+extension SpeechPipelineProfiles {
+    /// Convenience property for getting a profile for use by `SpeechPipelineBuilder`.
+    public var set: [SpeechProcessors]  {
```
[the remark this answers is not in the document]
noelweichbrodt: this is a bit of swift voodoo, `set` is a builtin.

Spokestack/Trace.swift
```diff
@@ -27,24 +27,12 @@ public struct Trace {
     /// - Parameter message: The debugging trace message.
     /// - Parameter delegate: The delegate that should receive the debugging trace message.
     /// - Parameter caller: The sender of the debugging trace message.
-    public static func trace(_ level: Trace.Level, config: SpeechConfiguration?, message: String, delegate: SpeechEventListener?, caller: Any) {
+    public static func trace(_ level: Trace.Level, message: String, config: SpeechConfiguration?, context: SpeechContext?, caller: Any) {
```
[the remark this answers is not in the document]
noelweichbrodt: callers may have optionals for these values, consolidate the handling of that here to avoid repetitious caller code.

Spokestack/Trace.swift
```diff
@@ -82,19 +70,23 @@ public struct Trace {
     /// - Parameter fileName: The name of the file that will be created/appended with the data.
     /// - Parameter delegate: The delegate that should receive the debugging trace message with the spit results.
     /// - Note: https://clojuredocs.org/clojure.core/spit
-    public static func spit(data: Data, fileName: String, delegate: SpeechEventListener, config: SpeechConfiguration?) {
+    public static func spit(data: Data, fileName: String, context: SpeechContext?, config: SpeechConfiguration?) {
```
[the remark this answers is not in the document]
noelweichbrodt: dumping tflite model posteriors

Spokestack/VADTrigger.swift
```diff
@@ -0,0 +1,40 @@
...
+    public func process(_ frame: Data) {
+        if self.context.isSpeech && !self.context.isActive {
+            self.context.isActive = true
+            self.configuration.delegateDispatchQueue.async {
+                self.context.listeners.forEach { listener in
+                    listener.didActivate()
+                }
+            }
+        } else if !self.context.isSpeech && self.context.isActive {
+//            self.context.isActive = false
```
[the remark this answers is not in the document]
noelweichbrodt: good catch, got abandoned during a refactor.

Spokestack/WebRTCVAD.swift
```diff
@@ -30,33 +30,46 @@ private var sampleRate32: Int32 = 16000
...
+    public func startStreaming() {}
     
-    /// Callback delegate for activation and error events.
-    public var delegate: VADDelegate?
+    public func stopStreaming() {}
+    
+    public init(_ configuration: SpeechConfiguration, context: SpeechContext) {
+        self.configuration = configuration
+        self.context = context
+        self.context.isSpeech = false
```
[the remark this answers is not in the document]
noelweichbrodt: this allows for a non-vad-using config to work without out of the box

Spokestack/WebRTCVAD.swift
```diff
@@ -30,33 +30,46 @@ private var sampleRate32: Int32 = 16000
...
+        } catch let error {
+            self.context.listeners.forEach({ listener in
+                listener.failure(speechError: error)
+            })
+        }
+    }
     
     deinit {
+        vad.deinitialize(count: 1)
         WebRtcVad_Free(vad.pointee)
```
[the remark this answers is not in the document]
noelweichbrodt: i've grown fond of my pointees

Spokestack/WebRTCVAD.swift
```diff
@@ -66,22 +79,26 @@ public class WebRTCVAD: NSObject {
         if errorCode != 0 { throw VADError.initialization("unable to create a WebRTCVAD struct, which returned error code \(errorCode)") }
         errorCode = WebRtcVad_Init(vad.pointee)
         if errorCode != 0 { throw VADError.initialization("unable to initialize WebRTCVAD, which returned error code \(errorCode)") }
-        errorCode = WebRtcVad_set_mode(vad.pointee, Int32(mode.rawValue))
+        errorCode = WebRtcVad_set_mode(vad.pointee, Int32(c.vadMode.rawValue))
         if errorCode != 0 {
-            WebRtcVad_Free(vad.pointee)
+            vad.pointee = nil
```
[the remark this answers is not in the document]
noelweichbrodt: this prevents a very special double-dealloc state, order is very purposeful.

SpokestackFrameworkExample/AppleWakewordViewController.swift
```diff
@@ -49,13 +49,12 @@ class AppleWakewordViewController: UIViewController {
...
-        c.tracing = Trace.Level.DEBUG
-        return SpeechPipeline(SpeechProcessors.appleSpeech.processor,
-                              speechConfiguration: c,
-                              speechDelegate: self,
-                              wakewordService: SpeechProcessors.appleWakeword.processor,
-                              pipelineDelegate: self)
+        return SpeechPipelineBuilder()
+            .setListener(self)
+            .useProfile(.appleWakewordAppleSpeech)
+            .setProperty("delegateDispatchQueue", "DispatchQueue.main")
```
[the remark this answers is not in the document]
noelweichbrodt: good catch, this was leftover from debugging

SpokestackFrameworkExample/AppleWakewordViewController.swift
```diff
@@ -49,13 +49,12 @@ class AppleWakewordViewController: UIViewController {
...
+            .setProperty("tracing", "Trace.Level.DEBUG")
```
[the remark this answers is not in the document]
noelweichbrodt: also a leftover from debugging. either actually works.

SpokestackTests/AppleSpeechRecognizerTest.swift
```diff
@@ -15,19 +15,44 @@ class AppleSpeechRecognizerTest: XCTestCase {
...
     /// startStreaming
     func testStartStopStreaming() {
-        let delegate = AppleSpeechRecognizerTestDelegate()
-        let asr = AppleSpeechRecognizer.sharedInstance
         let context = SpeechContext()
-        asr.configuration = SpeechConfiguration()
-        asr.delegate = delegate
-        asr.startStreaming(context: context)
+        let configuration = SpeechConfiguration()
+        let asr = AppleSpeechRecognizer(configuration, context: context)
```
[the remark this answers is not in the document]
noelweichbrodt: yep, although the mic may not be open long enough between `startStreaming` and `stopStreaming` for an actual request to get sent

SpokestackTests/AppleSpeechRecognizerTest.swift
```diff
@@ -15,19 +15,44 @@ class AppleSpeechRecognizerTest: XCTestCase {
...
+        context.isSpeech = true
+        asr.context = context
+        asr.startStreaming()
         XCTAssert(context.isActive)
-        asr.stopStreaming(context: context)
-        XCTAssert(!context.isActive)
+        XCTAssertFalse(delegate.didError)
+        asr.stopStreaming()
+        // asr does not set active
+        XCTAssertFalse(context.isActive)
```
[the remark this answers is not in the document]
noelweichbrodt: this is a redundant assertion for this test, good catch.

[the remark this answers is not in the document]
noelweichbrodt: stray comment from previous version, good catch

SpokestackTests/AppleSpeechRecognizerTest.swift
```diff
@@ -15,19 +15,44 @@ class AppleSpeechRecognizerTest: XCTestCase {
...
+        context.isActive = true
+        context.isSpeech = true
+        context.stageInstances = [asr]
+        asr.context = context
+        asr.startStreaming()
+        
+        asr.process(Frame.silence(frameWidth: 10, sampleRate: 8000))
+        asr.stopStreaming()
+        // asr does not set active
+        XCTAssertFalse(context.isActive)
```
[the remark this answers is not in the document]
noelweichbrodt: this is a redundant assertion for this test, good catch.

SpokestackTests/AudioControllerTest.swift
noelweichbrodt: it's a bit of defensiveness for tracking down whether a failure is due to test threading or due to a failed test assertion.

SpokestackTests/SpeechPipelineTest.swift
noelweichbrodt: i…don't remember. thanks for pointing that out.

SpokestackTests/SpeechPipelineTest.swift
noelweichbrodt: the latter

SpokestackTests/TFLiteWakewordRecognizerTest.swift
noelweichbrodt: because the initializer requires some state that `setUp` provides

SpokestackTests/TFLiteWakewordRecognizerTest.swift
noelweichbrodt: i believe what is there is considered idiomatic

noelweichbrodt:
After reading through the feedback one time, I think I can do a small refactor to move more control of stages and listeners to SpeechContext and remove some of the public surface for the API. Thanks for pushing that.

Spokestack/SpeechContext.swift
```diff
@@ -18,8 +19,87 @@ import Foundation
...
+                case .error:
+                    guard let e = self.error else {
+                        listener.failure(speechError: SpeechPipelineError.errorNotSet("A pipeline component attempted to send an error to SpeechContext's listeners without first setting the SpeechContext.error property."))
+                        return
+                    }
+                    listener.failure(speechError: e)
+                    self.error = .none
+                case .trace:
+                    guard let t = self.trace else {
+                        // swallow this error case because traces aren't important to the client.
```
space-pope, line 94: They might be (important). Best case here is that you're making app logging inconsistent based on what order the log listeners are attached—only the first listener will _ever_ get traces. I'd consider not clearing out the trace message immediately after sending. If you want it gone before the next event, wipe it after the loop has finished.

A bigger question here is whether it might make sense to adopt a more generic API for the listeners (like Android's single `onEvent` method) instead of having a separate method for each event. If this is the more idiomatic approach in Swift or iOS, so be it, but it does tend to make expansion a bit verbose.

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+
+        // stop
+        l.reset()
+        let didStopExpectation = expectation(description: "didStopExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.didStop.")
+        l.asyncExpectation = didStopExpectation
+        c.notifyListener(.stop)
+        wait(for: [didStopExpectation], timeout: 1)
+        XCTAssert(l.didDidStop)
+
+        // Activate
```
space-pope, line 72:
```suggestion
        // activate
```

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+        l.reset()
+        let didActivateExpectation = expectation(description: "didActivateExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.Activate.")
```
space-pope, line 74:
```suggestion
        let didActivateExpectation = expectation(description: "didActivateExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.activate.")
```

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+        l.asyncExpectation = didActivateExpectation
+        c.notifyListener(.activate)
+        wait(for: [didActivateExpectation], timeout: 1)
+        XCTAssert(l.activated)
+
+        // dectivate
```
space-pope, line 80:
```suggestion
        // deactivate
```

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+        l.reset()
+        let didDeactivateExpectation = expectation(description: "didDeactivateExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.dectivate.")
```
space-pope, line 82:
```suggestion
        let didDeactivateExpectation = expectation(description: "didDeactivateExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.deactivate.")
```

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+        l.reset()
+        // no error set still sends the event
+        let didErrorErrorExpectation = expectation(description: "didErrorErrorExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.error.")
+        l.asyncExpectation = didErrorErrorExpectation
+        c.notifyListener(.error)
+        wait(for: [didErrorErrorExpectation], timeout: 1)
+        // error set sends event
+        let didErrorExpectation = expectation(description: "didErrorExpectation fulfills when testNotifyListener calls SpeechContextTestDelegate.error.")
+        l.asyncExpectation = didErrorExpectation
+        c.notifyListener(.error)
```
space-pope, line 113: Doesn't look like you're actually setting an error here.

SpokestackTests/SpeechContextTest.swift
```diff
@@ -0,0 +1,233 @@
...
+    var asyncExpectation: XCTestExpectation?
+    
+    func reset() {
+        didDidInit = false
+        asyncExpectation = .none
+    }
+    
+    func didRecognize(_ result: SpeechContext) {
+        guard let _ = asyncExpectation else {
+            XCTFail("SpeechPipelineTestDelegate was not setup correctly. Missing XCTExpectation reference")
```
space-pope, line 150: Might be a little silly of me to make this sort of comment in a test, but this guard could be its own `validateSetup()` or something method, which would make the rest of the listener method implementations slightly cleaner.

noelweichbrodt:
This last commit allows for a certain amount of user extensibility of stages (`setStage`, `removeStage`) while reducing the API surface area of stages and simplifying the implementation of profiles by having `SpeechPipeline` manage stages.

noelweichbrodt:
Just realized I should have done an indexed for loop instead of the double reduce in the pipeline tests :)

Spokestack/SpeechContext.swift
```diff
@@ -18,8 +19,87 @@ import Foundation
     @objc public var isActive: Bool = false
     /// Speech detected indicator. Default to true for non-vad activation
     @objc public var isSpeech: Bool = true
-    /// `SpeechProcessor` instances that process audio frames from `AudioController`.
+    /// A set of `SpeechProcessor` instances that process audio frames from `AudioController`.
     public var stageInstances: [SpeechProcessor] = []
-    /// `SpeechEventListener`s that are sent `SpeechPipeline` events.
-    public var listeners: [SpeechEventListener] = []
+    /// A set of `SpeechEventListener`s that are sent `SpeechPipeline` events.
```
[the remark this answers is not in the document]
noelweichbrodt: I can’t implement Hashable in SpeechEventListener because of objc compatibility, so `Set` is unavailable for this use :( 
To save clients from themselves the observers will be an ordered set because I cannot imagine a usage where a client would deliberately register the same instance twice.

Spokestack/SpeechContext.swift
```diff
@@ -18,8 +19,87 @@ import Foundation
...
+    /// Removes all listeners.
+    @objc public func removeListeners() {
+        self.listeners = []
+    }
+
+    @objc public func setStage(_ stage: SpeechProcessor) {
+        
+    }
+
+    @objc public func removeStage(_ stage: SpeechProcessor) {
```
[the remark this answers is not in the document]
noelweichbrodt: I don't have a use case for why it's needed, just did it for completeness.

SpokestackTests/SpeechContextTest.swift
noelweichbrodt: Yeah, it's actually testing the guard clause fallback, need to add a happy path test too.

SpokestackTests/SpeechContextTest.swift
noelweichbrodt: This is all copypasta code for me, so I'm not worried about verboseness

Spokestack/SpeechPipeline.swift
```diff
@@ -81,27 +106,14 @@ import Dispatch
...
-                case .appleSpeech:
-                    return AppleSpeechRecognizer(self.configuration, context: self.context)
-                case .vadTrigger:
-                    return VADTrigger(self.configuration, context: self.context)
-                }
-            }()
-            self.context.stageInstances.append(stageInstance)
-        })
+        self.configuration.stages.forEach { stage in
+            self.stages.append(stage)
```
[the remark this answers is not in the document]
noelweichbrodt: good point

Spokestack/SpeechContext.swift
noelweichbrodt: dispatch for individual events is idiomatic in iOS (back to the message passing days), so I'll accept the verboseness in this case.

Spokestack/SpeechContext.swift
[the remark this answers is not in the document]
noelweichbrodt: It was clear, i just wasn't considering the case of a manually-activated pipeline.

Spokestack/SpeechContext.swift
```diff
@@ -56,40 +56,58 @@ import Foundation
...
-                }
+                }}
+            self.error = .none
+        case .trace:
+            if let t = self.trace {
+                self.listeners.forEach { listener in
+                    self.configuration.delegateDispatchQueue.async {
+                        listener.didTrace(t)
+                    }}
+                self.trace = .none
```
space-pope, line 104: If you really want to clear this message out (something that's not done in Android, and it actually just helped me catch a bug, almost by accident), you could still save yourself some code by moving the switch back where you had it and checking for the trace event after the loop's over, clearing it there.

SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -170,46 +170,46 @@ class SpeechPipelineBuilderTest: XCTestCase {
 
         // tflite
         delegate.asyncExpectation = didInit1Expectation
-        let p1 = SpeechPipelineBuilder()
+        let p1 = try! SpeechPipelineBuilder()
```
space-pope, line 173: Maybe a quick sad path test since profile is required now?

Spokestack/SpeechPipeline.swift
```diff
@@ -7,184 +7,195 @@
...
-        /// see previous comment
-        self.wakewordRecognizerService.delegate = self.speechDelegate
-        self.wakewordRecognizerService.configuration = speechConfiguration
-        
-        AudioController.sharedInstance.configuration = speechConfiguration
-        
-        self.pipelineDelegate = pipelineDelegate
-        AudioController.sharedInstance.pipelineDelegate = self.pipelineDelegate
+    /// - Parameter listeners: Delegate implementations of `SpeechEventListener` that receive speech pipeline events.
+    @objc public init(configuration: SpeechConfiguration, listeners: [SpeechEventListener]) {
```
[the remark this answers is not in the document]
space-pope: Looking at this holistically now, `profile` has become required in the builder, and this initializer sets up a pipeline with no stages (which makes it basically a no-op), so should it go away or have its arguments change such that it constructs a working pipeline?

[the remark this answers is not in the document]
noelweichbrodt: The public constructor is intended to support a 3rd-party-stages use, I can change the docstring to note that.

Spokestack/AppleSpeechRecognizer.swift
```diff
@@ -140,8 +140,8 @@ import Speech
                             a.confidence <= b.confidence }).first?.confidence ?? 0.0
                     strongSelf.context.transcript = r.bestTranscription.formattedString
                     strongSelf.context.confidence = confidence
-                    strongSelf.context.dispatch(.recognize)
                     strongSelf.vadFallWorker = DispatchWorkItem {[weak self] in
+                        self?.context.dispatch(.recognize)
```
space-pope, line 144: Putting this inside the vadFall task seems a little brittle to me. It seems better to either
a) check `r.isFinal` to determine whether to dispatch, or
b) set `shouldReportPartialResults` to `false` on the recognition request.

A note on b): we recently [started reporting partial results](https://github.com/spokestack/spokestack-android/pull/80) in the Android version.

noelweichbrodt: You'll love this: `isFinal` is never `true`. Been that way for years. Which is why `shouldReportPartialResults` must be `true`.

noelweichbrodt: Which is why this is all in my own custom vad fall timeout

space-pope: #$%@##%. Carry on.

Is `shouldReportPartialResults` similarly broken?

noelweichbrodt: Do you even have to ask? 😷

noelweichbrodt: to be fair, it's actually working and the only way to get a transcript, hence this other machinery to utilize it.

PR #87, opened 2020-08-14
restore objc interop for recognizers; fix double free bug in webtrcvad

noelweichbrodt:
The 11.0 release inadvertently removed some `@objc` attributes.

### DIFF
--- a/Spokestack/WebRTCVAD.swift
+++ b/Spokestack/WebRTCVAD.swift
@@ -55,7 +55,6 @@ private var frameBuffer: RingBuffer<Int16>!
     
     deinit {
         vad.deinitialize(count: 1)
-        WebRtcVad_Free(vad.pointee)
     }
     
     /// Creates and configures a new WebRTC VAD component.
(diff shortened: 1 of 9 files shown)


SpokestackTests/SpeechPipelineTest.swift
```diff
@@ -20,7 +20,7 @@ class SpeechPipelineTest: XCTestCase {
         let config = SpeechConfiguration()
 
         // successful init calls didInit
-        _ = SpeechPipeline(configuration: config, listeners: [delegate], stages: [])
+        let p = SpeechPipeline(configuration: config, listeners: [delegate], stages: [])
```
space-pope, line 23: necessary?

Spokestack/WebRTCVAD.swift
```diff
@@ -55,7 +55,6 @@ private var frameBuffer: RingBuffer<Int16>!
     
     deinit {
         vad.deinitialize(count: 1)
-        WebRtcVad_Free(vad.pointee)
```
space-pope, line 58: This change was intentional, right? Just checking; I know I've asked about these calls before.

noelweichbrodt: yeah, this is the "fix double free bug in webtrcvad" part of the pr. found another condition (caused by the line change you called out above) that causes a double free.

PR #89, opened 2020-08-17
SpeechContext is required in the SpeechPipeline stages constructor

noelweichbrodt:
Fixes a circular reference problem when a client does not use profiles + constructs stages (which is required when not using profiles). The pipeline creates its own `SpeechContext` instance, which will be different than the context used in stages. Since the pipeline now takes in the `SpeechContext` in this usage, there is no longer a need for the context to manage listeners, so `add`/`remove` for listeners has been rescoped to `internal`. The new test case is included to catch the bug that the circular constructor causes: a stage calls `dispatch` to a context with no listeners.


noelweichbrodt:
I agree—the `SpeechProcessor` interface should not require `context` as a public property, but as an argument in `start`/`stop`/`process`. 

That said, this PR is motivated by the tactical need to get the earlier pipeline refactor out, so I'll punt on that change for now.

PR #90, opened 2020-08-18
Fix tracing in pipeline builder and correct tracing bug in TFLiteWakewordRecognizer

noelweichbrodt:
- fix bug in pipeline builder that prevented tracing level from being set
- fix incorrect equality check for some TFLiteWakeword tracing

### DIFF
--- a/Spokestack/SpeechPipeline.swift
+++ b/Spokestack/SpeechPipeline.swift
@@ -161,7 +161,7 @@ import Dispatch
          .addListener(self)
          .setDelegateDispatchQueue(DispatchQueue.main)
          .useProfile(.tfLiteWakewordAppleSpeech)
-         .setProperty("tracing", ".PERF")
+         .setProperty("tracing", Trace.Level.PERF)
          .setProperty("detectModelPath", detectPath)
          .setProperty("encodeModelPath", encodePath)
          .setProperty("filterModelPath", filterPath)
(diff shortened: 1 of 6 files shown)


Spokestack/SpeechPipeline.swift
```diff
@@ -186,10 +186,19 @@ import Dispatch
     /// - SeeAlso: `SpeechConfiguration`
     /// - Parameters:
     ///   - key: Configuration property name
-    ///   - value: Configuration property name
+    ///   - value: Configuration property value
+    /// - Note: "tracing" key must have a value of `Trace.Level`, eg `Trace.Level.DEBUG`.
```
space-pope, line 190: Doesn't this change make for an otherwise unnecessary backflip during RN setup?

Spokestack/TFLiteWakewordRecognizer.swift
```diff
@@ -150,7 +150,7 @@ import TensorFlowLite
         
         // Tracing
         self.traceLevel = c.tracing
-        if self.traceLevel.rawValue >= Trace.Level.DEBUG.rawValue {
+        if self.traceLevel.rawValue <= Trace.Level.DEBUG.rawValue {
```
space-pope, line 153: There are a lot of fields here just for tracing that could probably be collapsed into a single data structure. Also, the trace level checks throughout the file oscillate between `< PERF` and `<= DEBUG`; it seems like you just want `== DEBUG` since it's the lowest level.

I think a bigger question might be that since the code for this is settled, is this file-based debugging of every piece of the wakeword code still helpful? Certainly some tracing for resets/activations is worth leaving in, but are the finer details useful for normal debugging?

Spokestack/SpeechPipeline.swift
noelweichbrodt: This change has no effect on the current RN setup.

Spokestack/TFLiteWakewordRecognizer.swift
noelweichbrodt: Good catch with the `< PERF` and `<= DEBUG` oscillations, that was crufty code.

This code is only settled for the current implementation. The implementation has already changed twice, with a third (multi-wakeword) in flight, so it is useful to keep the most intensive level of tracing messages for debugging purposes intact.

noelweichbrodt: `<= .DEBUG` preserves future compatibility if the tracing levels change.

space-pope: Yeah, I considered the possibility of changing levels, but that seemed highly unlikely. Fair point about the upcoming wakeword addition.

noelweichbrodt: I have thought that an `INTERNAL` level below `DEBUG` might be suitable for some of tracing messages in this class specifically.

PR #91, opened 2020-08-20
Fix double-sending recognize/deactivate events in AppleSpeechRecognizer

noelweichbrodt:
Also restore `timeout` event to AppleSpeechRecognizer.

### DIFF
--- a/Spokestack/AppleSpeechRecognizer.swift
+++ b/Spokestack/AppleSpeechRecognizer.swift
@@ -62,6 +62,7 @@ import Speech
         }
         self.audioEngine.prepare()
         self.wakeActiveMaxWorker = DispatchWorkItem {[weak self] in
+            self?.context.dispatch(.timeout)
             self?.deactivate()
         }
     }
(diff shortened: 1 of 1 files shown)


Spokestack/AppleSpeechRecognizer.swift
```diff
@@ -83,15 +84,18 @@ import Speech
...
+            self.active = false
+            self.context.isActive = false
+            self.recognitionTask?.finish()
+            self.recognitionTask = nil
+            self.recognitionRequest?.endAudio()
+            self.recognitionRequest = nil
+            self.vadFallWorker?.cancel()
+            self.wakeActiveMaxWorker?.cancel()
+            self.audioEngine.pause()
+            self.context.dispatch(.deactivate)
```
space-pope, line 97: This is a good reason for event dispatching to happen in the speech context. If I'm reading correctly, all calls other than the event dispatch (and maybe the audio engine pause?) should be idempotent, and if the context managed dispatching events on activity edge, you'd have nothing to worry about here.

PR #92, opened 2020-08-27
fix com.apple.coreaudio.avfaudio error -10868

noelweichbrodt:
This is all apple internal voodoo, there's not really a great way to test this fix other than smoke testing, which I have done extensively across two physical devices and in simulator.

### DIFF
--- a/Spokestack/AppleSpeechRecognizer.swift
+++ b/Spokestack/AppleSpeechRecognizer.swift
@@ -32,28 +32,31 @@ import Speech
     private var vadFallWorker: DispatchWorkItem?
     private var wakeActiveMaxWorker: DispatchWorkItem?
     private var active = false
+    private var bufferSize: Int
     
     // MARK: NSObject implementation
     
     deinit {
+        self.audioEngine.stop()
+        self.audioEngine.inputNode.removeTap(onBus: 0)
         speechRecognizer.delegate = nil
     }
     
     @objc public init(_ configuration: SpeechConfiguration, context: SpeechContext) {
         self.configuration = configuration
         self.context = context
+        self.bufferSize = 320 //(configuration.sampleRate / 1000) * configuration.frameWidth
         super.init()
     }
     
     // MARK: Private functions
     
     private func prepare() {
-        let bufferSize: Int = (self.configuration.sampleRate / 1000) * self.configuration.frameWidth
         self.audioEngine.inputNode.removeTap(onBus: 0) // a belt-and-suspenders approach to fixing https://github.com/wenkesj/react-native-voice/issues/46
         self.audioEngine.inputNode.installTap(
             onBus: 0,
             bufferSize: AVAudioFrameCount(bufferSize),
-            format: nil)
+            format: self.audioEngine.inputNode.inputFormat(forBus: 0))
         {[weak self] buffer, when in
             guard let strongSelf = self else {
                 return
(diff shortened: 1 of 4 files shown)


Spokestack/AppleSpeechRecognizer.swift
```diff
@@ -69,6 +72,8 @@ import Speech
     
     private func activate() {
         do {
+            self.configuration.tracing.rawValue <= Trace.Level.DEBUG.rawValue  ?
```
space-pope, line 75: So if tracing's set to debug, the audio engine won't actually start, but just dump info? Trying to parse the ternary here...this could stand another pass for readability.

Spokestack/AppleSpeechRecognizer.swift
```diff
@@ -32,28 +32,31 @@ import Speech
...
     deinit {
+        self.audioEngine.stop()
+        self.audioEngine.inputNode.removeTap(onBus: 0)
         speechRecognizer.delegate = nil
     }
     
     @objc public init(_ configuration: SpeechConfiguration, context: SpeechContext) {
         self.configuration = configuration
         self.context = context
+        self.bufferSize = 320 //(configuration.sampleRate / 1000) * configuration.frameWidth
```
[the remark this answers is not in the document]
noelweichbrodt: good catch, meant to refactor this part and got distracted

Spokestack/AppleSpeechRecognizer.swift
noelweichbrodt: the else part of the ternary is empty. the switch is only there because getting the debug information is costly so we don't want to do it in prod, and it's a ternary to make it all a one-liner so that it doesn't get in the way of the main logic.

space-pope: Whew, OK. I like to encapsulate the level checks in another class, though that gets messy in its own way.

I think just having an if statement would be less head-scratchy for readers—or you could even make it a separate fn with no extra fuss since it's not using any locals to dump the info. That might be the minimum distraction from the main logic.

Also: you prefix everything but `bufferSize` with `self.`.

noelweichbrodt: the level checks are done in another class, but in this case the trace message information is costly so a duplicate check beforehand makes sense.

Spokestack/SpeechConfiguration.swift
```diff
@@ -125,4 +125,6 @@ import Foundation
     @objc public var tracing: Trace.Level = Trace.Level.NONE
     /// Delegate events will be sent using the specified dispatch queue.
     @objc public var delegateDispatchQueue: DispatchQueue = DispatchQueue.global(qos: .userInitiated)
+    /// The dynamic size of the buffer in use by the `AudioEngine`.
+    internal var audioEngineBufferSize: UInt32 = 320
```
space-pope, line 129: You do document it as dynamic here, but shouldn't the actual value be dependent on the other config variables?

noelweichbrodt: no, it floats based on what the input format is. i don't need it any specific size where it's being used, it just needs to match across AudioEngine/AUUnits.

space-pope: Huh, OK. Apple 🤷

PR #96, opened 2020-09-15
add stopSpeaking api to tts

noelweichbrodt:
The hardest part was figuring out how to handle the additional button in the example project.

### DIFF
--- a/Spokestack/TextToSpeech.swift
+++ b/Spokestack/TextToSpeech.swift
@@ -111,6 +111,12 @@ private let apiQueue = DispatchQueue(label: TTSQueueName, qos: .userInitiated, a
         self.synthesize(input: input, success: play)
     }
     
+    /// Stops playback of the current synthesi result.
+    @objc public func stopSpeaking() -> Void {
+        self.player.pause()
+        self.finishPlayback()
+    }
+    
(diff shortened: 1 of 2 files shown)


Spokestack/TextToSpeech.swift
```diff
@@ -111,6 +111,12 @@ private let apiQueue = DispatchQueue(label: TTSQueueName, qos: .userInitiated, a
         self.synthesize(input: input, success: play)
     }
     
+    /// Stops playback of the current synthesi result.
```
space-pope, line 114:
```suggestion
    /// Stops playback of the current synthesis result.
```

space-pope (commented):
LGTM, with one question:

If a client calls `speak` twice in quick succession, is the audio for the second tts result queued? If so, does `stopSpeaking` have any implications for queued audio?

noelweichbrodt:
per line 108, each call to `speak` will replace Spokestack's `AVPlayer.currentItem` with the new synthesis result. It maintains no queue of results.

space-pope (approved):
Hm, I don't think that's necessarily the best approach for TTS (imagine a long prompt followed by a short one), but not an issue for this PR either way.

PR #97, opened 2020-09-16
asr components deliver partial transcript results via new optional partiallyRecognized api

noelweichbrodt:
Equivalent to https://github.com/spokestack/spokestack-android/pull/80/files

Due to objc interop requirements, I've utilized the swift/objc `optional` for protocols to implement the new `partiallyRecognized` delegate function. Existing clients will only need to opt-in the new function to utilize.


Spokestack/SpeechEventListener.swift
```diff
@@ -14,33 +14,37 @@ import Foundation
...
     /// The pipeline deactivate event. Occurs upon deactivation of speech recognition.  The pipeline remains inactive until activated again by either explicit activation or wakeword activation.
-    func didDeactivate() -> Void
+    @objc func didDeactivate() -> Void
     
     /// The pipeline recognized and transcribed speech.
     /// - Parameter result: The speech pipeline context, which contains the result.
-    func didRecognize(_ result: SpeechContext) -> Void
+    @objc func didRecognize(_ result: SpeechContext) -> Void
+    
+    /// The  pipeline recognized and transcripted an incomplete portion of the utterance.
```
space-pope, line 36:
```suggestion
    /// The pipeline recognized and transcribed a portion of an incomplete utterance.
```
Based on that description, isn't it more `didRecognizePartial` than `didPartiallyRecognize`?

Spokestack/SpokestackSpeechRecognizer.swift
```diff
@@ -106,8 +106,13 @@ import CryptoKit
     private func receive(result: (Result<URLSessionWebSocketTask.Message, Error>)) {
         self.handle(result, handleResult: { r in
             if let hypothesis = r.hypotheses.first {
-                self.context.confidence = hypothesis.confidence
-                self.context.transcript = hypothesis.transcript
+                if !hypothesis.transcript.isEmpty {
+                    if hypothesis.transcript != self.context.transcript {
```
space-pope, line 110: An extra string equality check feels like unnecessary overhead to me—both the chances of the system producing a bunch of identical transcripts in a row, and the cost of receiving duplicate partial results seem low.

space-pope (commented):
LGTM except for that one typo. Other comments are just suggestions.

Spokestack/SpokestackSpeechRecognizer.swift
noelweichbrodt: > the chances of the system producing a bunch of identical transcripts in a row
this is exactly why I had to do it. otherwise it's just walls of either empty or duplicated partial transcripts

Spokestack/SpeechEventListener.swift
noelweichbrodt: I can go with that

Spokestack/SpokestackSpeechRecognizer.swift
space-pope: Huh. Interesting. I don't remember many duplicates in my testing, but maybe I wasn't creative enough with my utterances.

space-pope (approved):

PR #98, opened 2020-09-23
Miscellaneous cleanup work

noelweichbrodt:
Nothing major here, just internal upgrades, removals, and test changes.


space-pope (approved):
I see no problems with your housekeeping acumen.

PR #101, opened 2020-10-07
minor adjustments to the Spokestack facade ergonomics

noelweichbrodt:
Based on writing the documentation, some minor improvements to using `Spokestack`. `Spokestack` has not appeared in a release yet, so there's still an opening to adjust without consequences.


space-pope (approved):
Anything that safely gets rid of `?`s LGTM.

PR #102, opened 2020-10-08
Configurable automatic asr utterance classification and optional transcript editing pre-classification

noelweichbrodt:
Since iOS already uses a global configuration object, automatic utterance classification uses for configuration. Otherwise should be equivalent to the Android implementation.

### DIFF
--- a/Spokestack/Spokestack.swift
+++ b/Spokestack/Spokestack.swift
@@ -30,7 +30,8 @@ import Foundation
     @objc public var configuration: SpeechConfiguration
     
     private var delegates: [SpokestackDelegate]
-    
+    private var editor: TranscriptEditor?
+
     /// This constructor is intended for use only by the `SpokestackBuilder`.
     /// - Parameters:
     ///   - delegates: Delegate implementations of `SpokestackDelegate` that receive Spokestack module events.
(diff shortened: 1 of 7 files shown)


Spokestack/Spokestack.swift
```diff
@@ -30,25 +30,41 @@ import Foundation
...
+    
+    /// A required function for `SpokestackDelegate` implementors
+    /// - Parameter error: An error sent from a Spokestack module.
+    public func failure(error: Error) {}
+    
+    /// An event receiver used to fulfill automatic classification configuration.
+    /// - Parameter result: Global state for the speech pipeline.
+    public func didRecognize(_ result: SpeechContext) {
+        self.configuration.automaticallyClassifyTranscript ?
+            self.nlu.classify(utterance: self.editor?.editTranscript(transcript: result.transcript) ?? result.transcript) : nil
```
space-pope, line 64: IMO this would be more readable if `utterance` were broken out as its own variable.

Spokestack/Spokestack.swift
```diff
@@ -128,6 +145,19 @@ import Foundation
         return self
     }
     
+    /// Sets a transcript editor used to alter ASR transcripts before they are classified by the NLU subsystem.
+    ///
+    /// If a transcript editor is set, registered listeners will still receive the `didRecognize` event from the speech pipeline with the unedited transcripts, but the editor will automatically run on those transcripts before the NLU module, operates on them. Thus, the `utterance` inside the `NLUResult` returned by classification will reflect the edited version of the transcript.
```
space-pope, line 150:
```suggestion
    /// If a transcript editor is set, registered listeners will still receive the `didRecognize` event from the speech pipeline with the unedited transcript, but the editor will automatically run on that transcript before the NLU module operates on it. Thus, the `utterance` inside the `NLUResult` returned by classification will reflect the edited version of the transcript.
```

SpokestackTests/SpokestackBuilderTest.swift
```diff
@@ -47,22 +63,66 @@ class SpokestackBuilderTest: XCTestCase {
...
+    
+    func stopStreaming() {}
+    
+    public init(_ context: SpeechContext, _ config: SpeechConfiguration) {
+        self.context = context
+        self.configuration = config
+    }
+    
+    func process(_ frame: Data) {
+        self.context.transcript = ".egabbac faed A"
```
space-pope, line 87: one heckuva asr you've got there

space-pope (approved):
LGTM; just had one potential readability improvement (and a doc thing).

Spokestack/Spokestack.swift
noelweichbrodt: I really miss `->>`

space-pope: Ditto. RIP.

PR #103, opened 2020-10-09
spokestack asr sends a timeout if the activation max ends without a transcript

noelweichbrodt:
The other asr already has this behavior.


space-pope (approved):
Looks a little strange to:

1. check for emptiness twice and
2. do one check with `isEmpty` and one with `!= ""`

but those are relatively minor gripes.

noelweichbrodt:
eh, good point, i missed the earlier if statement

PR #105, opened 2020-10-19
reorder profile enums for react native interop

noelweichbrodt:


space-pope (approved):
Seems a bit odd that the Apple profiles wouldn't all be clumped together, but nbd.

noelweichbrodt:
That's for ease of cross-platform indexing, since `appleWakeword` isn't cross-platform.

PR #106, opened 2020-10-29
Reorder AppleWakewordRecognizer ASR timeout dispatch to avoid forced unwrap

noelweichbrodt:
This only occurs  when apple wakeword is running for more than a minute.

fixes https://forum.spokestack.io/t/safely-stop-speechpipeline-on-ios/63


space-pope (approved):
I take it XCode won't let you write `self.dispatchWorker?.perform()`?

noelweichbrodt:
yeah, that evaluates to an optional, not a function

PR #107, opened 2020-11-16
fix bug ‘SFSpeechAudioBufferRecognitionRequest cannot be re-used’ in apple wakeword recognizer

noelweichbrodt:
Cf. https://forum.spokestack.io/t/occasional-crash-with-applewakeword/66

Tested using both sim and physical device. repro was only on physical device.

### DIFF
--- a/Spokestack.xcworkspace/contents.xcworkspacedata
+++ b/Spokestack.xcworkspace/contents.xcworkspacedata
@@ -2,9 +2,9 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Pods/Pods.xcodeproj">
+      location = "group:Spokestack.xcodeproj">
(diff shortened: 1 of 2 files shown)


space-pope (commented):
I think I get the idea here, but I'm a little confused why leaving the task open/running indefinitely would be feasible. It doesn't have any internal timeout or invalidation? I'd think you'd want to cancel as normal, then create a new task the next time `startRecognition` is called.

noelweichbrodt:
this allows the request object to be reused while cycling closing/opening the request. what you describe will also work, but it's just more thrashing to accomplish the same thing.

space-pope (approved):
Works for me. I'd expect the thrashing to be necessary, but great if it isn't.

PR #95, opened 2020-09-11
Add support for Spokestack cloud ASR

noelweichbrodt:
I did leave out frame chunking, but this is otherwise complete and comparable to the Android implementation.

PR #111, opened 2020-12-15
Keyword Recognizer

noelweichbrodt:
Adds support for on-device keyword recognition as an additional stage in the speech pipeline, with associated tests and reference implementation. The copypasta from on-device wakeword suggested a small number of internal improvements to both recognizers, hence the changes in `TFLiteWakewordRecognizer.swift`-associated files. Additionally, a set of mock and example keyword models are included for testing and reference use, respectively, in keeping with the practice established by the wakeword recognizer.

A small bug was fixed in the `RingBuffer` class to address allowing a `read` to continue after knowing it would fail.

Sibling of https://github.com/spokestack/spokestack-android/pull/95

### DIFF
--- a/Spokestack/Error.swift
+++ b/Spokestack/Error.swift
@@ -68,7 +68,8 @@ public enum VADError: Error, Equatable, LocalizedError {
     }
 }
 
-/// Errors thrown by implementors of the WakewordRecognizer protocol.
+/// Errors thrown by  command models.
+/// - SeeAlso: `TFLiteWakewordRecognizer`, `TFLiteKeywordRecognizer`
 public enum CommandModelError: Error, Equatable, LocalizedError {
     /// The command recognizer was unable to configure the recognizer model(s).
     case model(String)
--- a/Spokestack/Error.swift
+++ b/Spokestack/Error.swift
@@ -69,16 +69,16 @@ public enum VADError: Error, Equatable, LocalizedError {
 }
 
 /// Errors thrown by implementors of the WakewordRecognizer protocol.
-public enum WakewordModelError: Error, Equatable, LocalizedError {
-    /// The WakewordRecognizer was unable to configure the recognizer model(s).
+public enum CommandModelError: Error, Equatable, LocalizedError {
+    /// The command recognizer was unable to configure the recognizer model(s).
     case model(String)
-    /// The WakewordRecognizer encountered an error during the processing of the audio frame.
+    /// The rcommand ecognizer encountered an error during the processing of the audio frame.
     case process(String)
-    /// The WakewordRecognizer encountered an error during the configuration or running of the filter model.
+    /// The command recognizer encountered an error during the configuration or running of the filter model.
     case filter(String)
-    /// The WakewordRecognizer encountered an error during the configuration or running of the encode model.
+    /// The command recognizer encountered an error during the configuration or running of the encode model.
     case encode(String)
-    /// The WakewordRecognizer encountered an error during the configuration or running of the detect model.
+    /// The command recognizer encountered an error during the configuration or running of the detect model.
     case detect(String)
     
     /// `LocalizedError` implementation so that `localizedDescription` isn't an enum index.
--- a/Spokestack/Error.swift
+++ b/Spokestack/Error.swift
@@ -73,7 +73,7 @@ public enum VADError: Error, Equatable, LocalizedError {
 public enum CommandModelError: Error, Equatable, LocalizedError {
     /// The command recognizer was unable to configure the recognizer model(s).
     case model(String)
-    /// The rcommand ecognizer encountered an error during the processing of the audio frame.
+    /// The command recognizer encountered an error during the processing of the audio frame.
     case process(String)
     /// The command recognizer encountered an error during the configuration or running of the filter model.
     case filter(String)
--- a/Spokestack/RingBuffer.swift
+++ b/Spokestack/RingBuffer.swift
@@ -10,7 +10,7 @@ import Foundation
 
 /// A simple circular buffer of values.
 final class RingBuffer <T> {
-
+    
     // MARK: Public (properties)
     
     /// The number of empty spaces remaining in the RingBuffer until it's full.
@@ -81,11 +81,10 @@ final class RingBuffer <T> {
     func read() throws -> T {
         if self.isEmpty {
             throw RingBufferStateError.illegalState(message: "ring buffer is empty")
-        } else {
-            let value: T = self.data[self.rpos]
-            self.rpos = self.pos(self.rpos + 1)
-            return value
         }
+        let value: T = self.data[self.rpos]
+        self.rpos = self.pos(self.rpos + 1)
+        return value
     }
     
     /// Writes the next value to the buffer.
--- a/Spokestack/RingBuffer.swift
+++ b/Spokestack/RingBuffer.swift
@@ -81,10 +81,11 @@ final class RingBuffer <T> {
     func read() throws -> T {
         if self.isEmpty {
             throw RingBufferStateError.illegalState(message: "ring buffer is empty")
+        } else {
+            let value: T = self.data[self.rpos]
+            self.rpos = self.pos(self.rpos + 1)
+            return value
         }
-        let value: T = self.data[self.rpos]
-        self.rpos = self.pos(self.rpos + 1)
-        return value
     }
     
     /// Writes the next value to the buffer.
--- a/Spokestack/SpeechConfiguration.swift
+++ b/Spokestack/SpeechConfiguration.swift
@@ -21,9 +21,11 @@ import Foundation
     public var fftWindowType: SignalProcessing.FFTWindowType = .hann
     /// The desired linear Root Mean Squared (RMS) signal energy, which is used for signal normalization and should be tuned to the RMS target used during wakeword model training.
     /// - SeeAlso: `TFLiteWakewordRecognizer`
+    @available(*, deprecated, message: "RMS normalization is no longer used during wakeword recognition.")
     @objc public var rmsTarget: Float = 0.08
     /// The Exponentially Weighted Moving Average (EWMA) update rate for the current  Root Mean Squared (RMS) signal energy (0 for no RMS normalization).
     /// - SeeAlso: `TFLiteWakewordRecognizer`
+    @available(*, deprecated, message: "RMS normalization is no longer used during wakeword recognition.")
     @objc public var rmsAlpha: Float = 0.0
     /// The size of the signal window used to calculate the STFT, in number of samples - should be a power of 2 for maximum efficiency.
     /// - SeeAlso: `TFLiteWakewordRecognizer`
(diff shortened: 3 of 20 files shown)


Spokestack/RingBuffer.swift
```diff
@@ -81,10 +81,11 @@ final class RingBuffer <T> {
     func read() throws -> T {
         if self.isEmpty {
             throw RingBufferStateError.illegalState(message: "ring buffer is empty")
+        } else {
+            let value: T = self.data[self.rpos]
+            self.rpos = self.pos(self.rpos + 1)
+            return value
         }
-        let value: T = self.data[self.rpos]
```
brentspell, line 85: I'm wondering how you reached this line of code before you made this change if self.isEmpty returned true, and if swift exceptions really work like this, you have the same bug in your `write` function on the `self.isFull` check

Spokestack/SpeechConfiguration.swift
```diff
@@ -131,4 +131,62 @@ import Foundation
...
+    /// - SeeAlso: `TFLiteKeywordRecognizer`
+    @objc public var keywordDetectModelPath: String = "KeywordDetect.tflite"
+    /// The threshold of the keyword recognizer's posterior output, above which the keyword recognizer emits a recognition event for the most probable keyword.
+    /// - SeeAlso: `TFLiteKeywordRecognizer`
+    @objc public var keywordThreshold: Float = 0.5
+    /// A comma-separated list of keywords to recognize.
+    /// - Remark: ex: "yes,no"
+    /// - Warning: cannot contain spaces
+    /// - SeeAlso: `TFLiteKeywordRecognizer`
+    @objc public var keywords: String = "bed,bird,cat,dog,down,eight,five,four,go,happy,house,left,marvin,nine,no,off,on,one,right,seven,sheila,six,stop,three,tree,two,up,wow,yes,zero"
```
brentspell, line 162: this parameter should not have a default value - it should always be provided explicitly by the client

Spokestack/SpeechConfiguration.swift
```diff
@@ -131,4 +131,62 @@ import Foundation
...
+    /// The name of the window function to apply to each audio frame before calculating the STFT.
+    /// - Remark: Currently the "hann" window is supported.
+    /// - SeeAlso: `TFLiteWakewordRecognizer`
+    public var keywordFFTWindowType: SignalProcessing.FFTWindowType = .hann
+    /// The desired linear Root Mean Squared (RMS) signal energy, which is used for signal normalization and should be tuned to the RMS target used during wakeword model training.
+    /// - SeeAlso: `TFLiteWakewordRecognizer`
+    @objc public var keywordRMSTarget: Float = 0.08
```
brentspell, line 169: we don't do rms normalization for keyword models, and the feature is deprecated for wakeword models, in favor of automatic gain control

Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -0,0 +1,373 @@
...
+        self.fftWindow = SignalProcessing.fftWindowDispatch(windowType: c.keywordFFTWindowType, windowLength: c.keywordFFTWindowSize)
+        self.fft = FFT(c.keywordFFTWindowSize)
+        self.hopLength = c.keywordFFTHopLength * c.sampleRate / 1000
+        let melLength: Int = c.keywordMelFrameLength * c.sampleRate / 1000 / self.hopLength
+        self.frameWindow = RingBuffer(melLength * c.keywordMelFrameWidth, repeating: 0.0)
+        self.encodeState = RingBuffer(c.keywordEncodeWidth, repeating: 0.0)
+        self.encodeState.fill(0.0) // fill now because the encoded state is used to feed-forward
+        let encodeLength = c.keywordEncodeLength * c.sampleRate / 1000 / self.hopLength
+        self.encodeWindow = RingBuffer(encodeLength * c.keywordEncodeWidth, repeating: -1.0)
+        self.classes = self.configuration.keywords.components(separatedBy: ",")
```
brentspell, line 79: the length of the class list should be validated against the detector model output dimension, especially if you insist on keeping that silly default class list configuration

Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -0,0 +1,373 @@
...
+                self.sampleCollector?.append(sample)
+            }
+            
+            // Process the sample
+            // - write it to the sample sliding window
+            // - run the remainder of the detection pipleline if speech
+            // - advance the sample sliding window
+            try self.sampleWindow.write(sample)
+            if self.sampleWindow.isFull {
+                try self.analyze()
```
brentspell, line 151: the analyzer and filter/encoder models should only be invoked if the pipeline is active
https://github.com/spokestack/spokestack-android/blob/master/src/main/java/io/spokestack/spokestack/asr/KeywordRecognizer.java#L347
this protects the encoder model from non-speech or wakeword audio, for which the outputs are undefined

SpokestackTests/TFLiteKeywordRecognizerTest.swift
```diff
@@ -0,0 +1,197 @@
...
+        self.recognizer?.startStreaming()
+        let recognizeExpectation = XCTestExpectation(description: "process without failure.")
+        let timeoutExpectation = XCTestExpectation(description: "process without failure.")
+        let failureExpectation = XCTestExpectation(description: "process without failure.")
+        self.delegate.asyncExpectation = failureExpectation
+        self.delegate.didRecognizeExepctation = recognizeExpectation
+        self.delegate.didTimeoutExpectation = timeoutExpectation
+        
+        // timeout
+        for _ in 0...1 {
```
space-pope, line 87: Isn't this just one call?

SpokestackTests/TFLiteKeywordRecognizerTest.swift
```diff
@@ -0,0 +1,197 @@
...
+        self.context?.isActive = false
+        wait(for: [timeoutExpectation], timeout: 1)
+        XCTAssert(self.delegate.timedOut)
+        XCTAssertFalse(self.delegate.didError)
+        XCTAssertFalse(self.delegate.recognized)
+        
+        // don't process if the pipeline isn't active
+        delegate.reset()
+        self.recognizer?.startStreaming()
+        self.recognizer?.context.isSpeech = true
```
space-pope, line 99: Why set to true then immediately to false?

SpokestackTests/TFLiteKeywordRecognizerTest.swift
```diff
@@ -0,0 +1,197 @@
...
+    var recognized: Bool = false
+    var timedOut: Bool = false
+    var asyncExpectation: XCTestExpectation?
+    var didRecognizeExepctation: XCTestExpectation?
+    var didTimeoutExpectation: XCTestExpectation?
+    
+    func failure(error: Error) {
+        print(error)
+        guard let _ = asyncExpectation else {
+            XCTFail("TFLiteKeywordecognizerTestDelegate was not setup correctly. Missing XCTExpectation reference")
```
space-pope, line 159:
```suggestion
            XCTFail("TFLiteKeywordRecognizerTestDelegate was not setup correctly. Missing XCTExpectation reference")
```

Spokestack/RingBuffer.swift
space-pope: Yeah, that's...weird.

space-pope (commented):
I don't see anything wrong here, just some cleanup opportunity.

Spokestack/RingBuffer.swift
noelweichbrodt: I'm unable to replicate the magical thinking that led to this change.

Spokestack/SpeechConfiguration.swift
noelweichbrodt: objc compatibility requires all parameters to have default values. This one can be set to an empty string to avoid confusion.

Spokestack/SpeechConfiguration.swift
noelweichbrodt: missed that change

Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -0,0 +1,373 @@
...
+        // Tracing
+        if c.tracing.rawValue <= Trace.Level.DEBUG.rawValue {
+            self.sampleCollector = []
+            self.fftFrameCollector = ""
+            self.filterCollector = []
+            self.encodeCollector = []
+        }
+        
+        // tensorflow model initialization
+        do {
```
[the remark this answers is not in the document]
noelweichbrodt: The model init and the configuration init are interrelated, so I'll leave them together. Model type errors are a simple convenience enum, a single error type would still need to specify model type in the error message anyways.

Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -0,0 +1,373 @@
...
+        
+        // filter() will utilize the stft spectrogram as input for the filter model
+        try self.filter()
+    }
+    
+    private func filter() throws -> Void {
+        // utilize the stft spectrogram as input for the filter model
+        do {
+            guard let model = self.filterModel else {
+                throw CommandModelError.filter("model was not initialized")
```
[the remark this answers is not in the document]
noelweichbrodt: the initialization error could be caught and ignored, so redundancy is helpful here

Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -0,0 +1,373 @@
...
+            guard let strongSelf = self else { return }
+            if strongSelf.context.isActive {
+                do {
+                    try strongSelf.sample(frame)
+                } catch let error {
+                    strongSelf.context.dispatch { $0.failure(error: error) }
+                }
+                // sample every frame while active
+                if !strongSelf.isActive {
+                    strongSelf.isActive = true
```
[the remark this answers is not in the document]
noelweichbrodt: the sample is taken in the preceding block

SpokestackTests/TFLiteKeywordRecognizerTest.swift
noelweichbrodt: two calls

SpokestackTests/TFLiteKeywordRecognizerTest.swift
noelweichbrodt: this is a test of the vad edge logic

Spokestack/Error.swift
```diff
@@ -69,16 +69,16 @@ public enum VADError: Error, Equatable, LocalizedError {
 }
 
 /// Errors thrown by implementors of the WakewordRecognizer protocol.
```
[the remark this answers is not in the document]
noelweichbrodt: good catch, that's a stale comment.

Spokestack/TFLiteKeywordRecognizer.swift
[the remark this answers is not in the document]
space-pope: I just meant something along the lines of a `loadModel` fn that encapsulates the if/else and error throwing; it'd be called once for each model in this `do` block.

Spokestack/TFLiteKeywordRecognizer.swift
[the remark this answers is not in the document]
space-pope: Any error can be caught and ignored :)

space-pope (approved):

Issue #110, opened 2020-12-09
AudioController memory build up


xaphod:
I see constant linear memory growth in XCode Memory.

This fixes it for me: in AudioController.swift, wrap contents of `func recordingCallback` with `autoreleasepool { }`

noelweichbrodt:
Fixed by #112

Issue #109, opened 2020-12-08
Long pauses (hangs) on pipeline.stop()


xaphod:
Disclaimer: this is 9.0.1 (from May 6 2020) because I support iOS 12.

I'm occasionally seeing long pauses / hangs on calling .stop(), with errors shown like this one:
`[aurioc] AURemoteIO.cpp:1639:Stop: AURemoteIO::Stop: error 268451843 calling TerminateOwnIOThread (port 104967)`
This seems to happen on iOS 12 and iOS 14 (errors are different) with similar frequency.

The impact is minimized when I call stop off the main thread, ie. `DispatchQueue.global().async { self.pipeline?.stop() }`, because at least none of the UI work is waiting for it.

2 questions:
1) did this issue get solved in later releases somehow - any tips? I have already forked, am happy to develop fixes manually etc if I have a direction to go in
2) what's the recommended threading model for calling start/stop on the speech pipeline? Currently i'm calling start() from main thread, stop() off main thread. Not sure that's wise...

Many thanks, enjoying spoke stack so far (day 3)

noelweichbrodt:
Hi Tim, glad to see you're getting further and enjoying Spokestack!

> did this issue get solved in later releases somehow - any tips? I have already forked, am happy to develop fixes manually etc if I have a direction to go in

There have been a number of bugs that have been fixed wrt threading and pipeline stop/start since v9.0.1. 

Assuming that you're just using Spokestack for the speech pipeline component and not TTS or NLU, I think it'd be easier to just update to the latest version and then rip out the stuff that requires iOS 13 (iir it's mainly TTS and NLU using `Combine` and `Crypto` libs, and Spokestack's cloud ASR). Additionally, that allows for your fork to benefit more easily from upstream improvements in the future.

> what's the recommended threading model for calling start/stop on the speech pipeline? Currently i'm calling start() from main thread, stop() off main thread. Not sure that's wise...

We recommend the pipeline to be run off a thread with a QoS of at least `.userInitiated`, but not on the main UI thread.

xaphod:
Thanks for the quick reply.
You were right, it was very easy to whip out the machete and get 14.0.3 working on iOS12, thanks.

First run through, I immediately hit the same hang & error again: 
`2020-12-08 16:16:05.948892-0500 booth[7564:6638313] [aurioc] AURemoteIO.cpp:1639:Stop: AURemoteIO::Stop: error 268451843 calling TerminateOwnIOThread (port 69403)`

[...]

PR #112, opened 2020-12-23
Apple recognizer concurrency fixes

noelweichbrodt:
Suggested by https://github.com/spokestack/spokestack-ios/compare/14.0.3...xaphod:xaphod/14.0.3 as fixes for #109 & #110.

Memory usage is now constant, and repeated simulator runs suggest the concurrency issues wrt `AudioEngine` are no longer impacting usage.

Also, some minor code changes and comments to help future debugging in `TextToSpeech`

### DIFF
--- a/Spokestack/AppleSpeechRecognizer.swift
+++ b/Spokestack/AppleSpeechRecognizer.swift
@@ -12,7 +12,7 @@ import Speech
 /**
  This pipeline component uses the Apple `SFSpeech` API to stream audio samples for speech recognition.
  
- Once speech pipeline coordination via `startStreaming` is received, the recognizer begins streaming buffered frames to the Apple ASR API for recognition. Once speech pipeline coordination via `stopStreaming` is received, or when the Apple ASR API indicates a completed speech event, the recognizer completes the API request and calls the `SpeechEventListener` delegate's `didRecognize` event with the updated global speech context (including the audio transcript and confidence).
+ Once speech pipeline coordination via `startStreaming` is received, the recognizer begins streaming buffered frames to the Apple ASR API for recognition. Once speech pipeline coordination via `stopStreaming` is received, or when the Apple ASR API indicates a completed speech event, the recognizer completes the API request and either sends a `timeout` or `didRecognize` event with the updated global speech context (including the speech transcript and confidence).
  */
 @objc public class AppleSpeechRecognizer: NSObject {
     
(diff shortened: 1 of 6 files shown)


Spokestack/AppleSpeechRecognizer.swift
```diff
@@ -36,6 +39,10 @@ import Speech
     // MARK: NSObject implementation
     
     deinit {
+        if recognizingSemaphore.wait(timeout: .now() + self.configuration.semaphoreTimeout) == .timedOut {
+            self.context.dispatch { $0.failure(error: SpeechPipelineError.illegalState("Could not start recognizer, timed out waiting on another thread.")) }
```
space-pope, line 43: This is a long error message to be copy pasta'd everywhere.

space-pope (commented):
Pretty sure I get what we're doing here, but I have a lingering concern: `AppleSpeechRecognizer` and `AppleWakewordRecognizer` use the _same_ external resources to do their thing, and they both rely on a thread pool (of unknown size, afaik) to submit their tasks. As such, in order to eliminate a potential race condition between the two, they'd have to share a semaphore rather than construct their own.

noelweichbrodt:
If you're talking about `SFSpeechAudioBufferRecognitionRequest` or `AVAudioEngine` wrt "use the same external resources to do their thing", then I see what you're concerned about but don't share it. `AVAudioEngine` is a resource graph that may be thought of as a multiplexed audio manager, so each instance operates independently by design. `SFSpeechAudioBufferRecognitionRequest` is just a single request that is expected to operate independently, and is fed audio from `AVAudioEngine`, so multiple requests held by different places may compete for resources, but should not be managed (outside of official guidance I haven't seen) by their instantiators.

Spokestack/AppleSpeechRecognizer.swift
noelweichbrodt: error message is pretty short, error handling is rather officious

space-pope:
I mean that the device has a single microphone that these classes have to compete for, but the `AVAudioEngine` abstraction does make working with it a little different than in Android-land. Since you're not managing the microphone directly, I'd wager there's a good chance of the ASR losing a frame (or a few) on the activation edge, but that's probably not a big deal.

space-pope (approved):

noelweichbrodt:
Yeah, there's exactly zero guarantees for real-time delivery of frames, it's all best-effort.

PR #114, opened 2021-02-09
reduce keyword recognizer receptive field

noelweichbrodt:
removal of the matchboxnet convolutions in the keyword models mean a smaller frame length and slightly longer encoding length. sibling of https://github.com/spokestack/spokestack-android/pull/129

also took the opportunity to improve a keyword model size check and error message.


brentspell (approved):
lgtm

PR #117, opened 2021-05-12
feature: keyword recognizer uses json metadata for classes if available

noelweichbrodt:


Spokestack/TFLiteKeywordRecognizer.swift
```diff
@@ -72,7 +72,16 @@ import TensorFlowLite
         self.encodeState.fill(0.0) // fill now because the encoded state is used to feed-forward
         let encodeLength = c.keywordEncodeLength * c.sampleRate / 1000 / self.hopLength
         self.encodeWindow = RingBuffer(encodeLength * c.keywordEncodeWidth, repeating: -1.0)
-        self.classes = self.configuration.keywords.components(separatedBy: ",")
+        do {
+            let keywords = try CommandModelMeta(configuration)
+            self.classes = keywords.model.classes.flatMap({ c in
```
space-pope, line 77: We don't want the utterance text here, just the class names—the model outputs will correspond to training classes, not individual utterances that resolve to those classes.

space-pope (changes requested):
There's an easy-to-make misunderstanding about the metadata here, noted below. I'd double-check the test metadata itself, but GitHub won't show it to me since it's tracked by git-lfs (I only do that for the model files themselves, fwiw).

Spokestack/TFLiteKeywordRecognizer.swift
noelweichbrodt: 🤦  good catch, thanks

space-pope (approved):
LGTM 👍🏼

